PR #2, opened 2020-09-26
Comprehensive improvements

SerafimArts:
Added features:
 - Basic options for Activity execution
 - Query and Signal handlers support
 - Basic PSR-3 implementation
 - Separated workflow executor
 - Complete generation of attribute-based declarations

Fixed bugs:
 - Incorrect answer when resolving a promise
 - Memory leaks when resolving deferreds

Well and something else...

### DIFF
--- a/example/Workflow/PizzaDelivery.php
+++ b/example/Workflow/PizzaDelivery.php
@@ -11,42 +11,40 @@ declare(strict_types=1);
 
 namespace App\Workflow;
 
+use Temporal\Client\Meta\QueryMethod;
+use Temporal\Client\Meta\SignalMethod;
 use Temporal\Client\Meta\WorkflowMethod;
 use Temporal\Client\Runtime\WorkflowContextInterface;
 
 class PizzaDelivery
 {
     /**
-     * Promise-based example:
-     *
-     * <code>
-     *  $context->executeActivity('ExecuteActivity')
-     *      ->then(function ($result) use ($context) {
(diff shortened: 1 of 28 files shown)


example/Workflow/PizzaDelivery.php
```diff
@@ -11,39 +11,40 @@
...
+        $b = yield $context->executeActivity('B');
 
-        return 42;
+        return [$a, $b];
+    }
+
+    /**
+     * @QueryMethod()
+     */
+    #[QueryMethod]
```
wolfy-j, line 36: is it 8 ready?

example/client.php
```diff
@@ -29,18 +32,15 @@
 //
 $worker = new Worker($transport, $loop);
 
+$worker->setLogger($logger);
```
wolfy-j, line 35: we will need per workflow logger as well, with replaying capabilities

wolfy-j (approved):
Crazy amount of work!

example/Workflow/PizzaDelivery.php
SerafimArts: Yep. You can choose one annotations type. 

Here it is duplicated in different syntax for the example.

example/client.php
SerafimArts: It should be located somewhere here: https://github.com/temporalio/php-sdk/blob/develop/src/Runtime/Workflow/Executor.php#L79

PR #19, opened 2020-12-29
- added signal queue

wolfy-j:


src/Internal/Queue/SignalQueue.php
```diff
@@ -0,0 +1,54 @@
...
+     * @param array $args
+     */
+    public function push(string $signal, array $args): void
+    {
+        if (isset($this->consumers[$signal])) {
+            ($this->consumers[$signal])($args);
+            return;
+        }
+
+        if (!isset($this->queue[$signal])) {
```
SerafimArts, line 21: `if` + `else` can be replaced to:

```php
$this->queue[$signal][] = $args;

if (isset($this->queue[$signal])) {
    $this->flush($signal);
}
```

PR #27, opened 2021-01-05
Add child workflow stub/proxy impl

SerafimArts:


src/DataConverter/Payload.php
```diff
@@ -50,6 +51,28 @@ public static function createRaw(array $metadata, string $data = null): Payload
         return $payload;
     }
 
+    /**
```
wolfy-j, line 54: to be optimized after merge

wolfy-j (approved):

PR #58, opened 2021-01-31
Usability improvements

wolfy-j:
- Shorter stack traces
- Wrap generators in coroutines
- Always cast milliseconds to int
- Scope won't close with pending Promises
- Wrap search attributes and memo into Payload


src/Client/WorkflowOptions.php
```diff
@@ -149,19 +150,21 @@ public function __construct()
      */
     public function mergeWith(MethodRetry $retry = null, CronSchedule $cron = null): self
     {
```
SerafimArts, line 152: what has changed in general?

src/Internal/Support/StackRenderer.php
```diff
@@ -11,6 +11,24 @@
 
 class StackRenderer
 {
+    /**
+     * Sets files and prefixes to be ignored from the stack trace.
+     *
+     * @var array
+     */
+    private static array $ignorePaths = [
+        'temporal/sdk/src/Internal/',
```
SerafimArts, line 20: Maybe the namespace makes sense? Instead of path i mean.

src/Internal/Marshaller/Type/DateIntervalType.php
```diff
@@ -53,10 +53,10 @@ public function serialize($value): int
         $method = 'total' . \ucfirst($this->format);
 
         if ($this->format === DateInterval::FORMAT_NANOSECONDS) {
-            return DateInterval::parse($value, $this->format)->totalMicroseconds * 1000;
+            return (int)(DateInterval::parse($value, $this->format)->totalMicroseconds * 1000);
         }
 
-        return DateInterval::parse($value, $this->format)->$method;
+        return (int)(DateInterval::parse($value, $this->format)->$method);
```
SerafimArts, line 59: same

src/Internal/Marshaller/Type/DateIntervalType.php
```diff
@@ -53,10 +53,10 @@ public function serialize($value): int
         $method = 'total' . \ucfirst($this->format);
 
         if ($this->format === DateInterval::FORMAT_NANOSECONDS) {
-            return DateInterval::parse($value, $this->format)->totalMicroseconds * 1000;
+            return (int)(DateInterval::parse($value, $this->format)->totalMicroseconds * 1000);
```
SerafimArts, line 56: Or maybe int64 overflow just happens? In this case, the conversion to the float occurs...

src/Worker/Transport/Codec/ProtoCodec/Decoder.php
```diff
@@ -70,7 +70,12 @@ private function parseRequest(Message $msg): RequestInterface
             $payloads = EncodedValues::fromPayloads($msg->getPayloads(), $this->converter);
         }
 
-        return new Request($msg->getCommand(), json_decode($msg->getOptions(), true), $payloads, $msg->getId());
+        return new Request(
+            $msg->getCommand(),
+            json_decode($msg->getOptions(), true),
```
SerafimArts, line 75: Missing `JSON_THROW_ON_ERROR`

src/Internal/Marshaller/Type/DateIntervalType.php
wolfy-j: Can't be the case, the test case where I got it was under 10000ms.

src/Internal/Support/StackRenderer.php
wolfy-j: Does not handle Closures.

SerafimArts (approved):
LGTM

PR #59, opened 2021-02-01
Сlient polishing

wolfy-j:
- Minor CS in Client
- Bump stability to Beta


src/Client/ActivityCompletionClientInterface.php
```diff
@@ -11,13 +11,11 @@
...
 namespace Temporal\Client;
 
-use Temporal\Exception\Client\ActivityCanceledException;
-
 /**
  * Used to complete asynchronously activities that called {@link
  * ActivityContext->doNotCompleteOnReturn()}.
  *
- * <p>Use {@link WorkflowClient#newActivityCompletionClient()} to create an instance.
+ * <p>Use {@link WorkflowClient->newActivityCompletionClient()} to create an instance.
```
SerafimArts, line 18: - `<p>` should be removed
- `@link` should be replace to `@see`

composer.json
```diff
@@ -81,6 +81,6 @@
     "config": {
         "sort-packages": true
     },
-    "minimum-stability": "dev",
+    "minimum-stability": "beta",
```
SerafimArts, line 84: doesnt make sence

SerafimArts (approved):

PR #87, opened 2021-03-03
Feature/no fatal on cleanup

wolfy-j:


SerafimArts:
This is a case for desynchronizing the state of the client and server.

Getting rid of the exception will correct the symptomatology of the error, but it will not solve the problem.

Are you sure you need it?

wolfy-j:
The only other alternative is to localize this issue and fail the specific workflow. I think we still have to seek the root cause of the de-sync (which seems to happen only under heavy load and reloading workflows from and into the cache in the cycle). However, now the error kills the whole worker instead of just a single workflow.

SerafimArts:
Then maybe add `@trigger_error('...')` to catch desynchronization and debug this cases in production? This will not be displayed anywhere, but these triggers can be caught in the future.

Issue #95, opened 2021-04-05
[Bug] Non-retryable exceptions list in RetryOptions does not work as expected


kozlice:
**Describe the bug**
Non-retryable exceptions list in RetryOptions does not work as expected.

**To Reproduce**
Type hints (`@psalm-type ExceptionsList = array<class-string<\Throwable>>`) and documentation (https://docs.temporal.io/docs/php-retries) tell us to provide a list of class names:
```php
$retryOptions = RetryOptions::new()
    ->withNonRetryableExceptions([
        \InvalidArgumentException::class,
        \LogicException::class,
    ])
    ->withMaximumAttempts(3)
    ->withInitialInterval(new \DateInterval('PT5S'))
    ->withBackoffCoefficient(1)
;
```
Then I throw `RuntimeException` inside an activity method, which is not on the list. However, instead of an expected retry I saw this in Temporal web UI:
```
PanicError: assert(Assert::valuesInstanceOf($exceptions, \Throwable::class)) (type: AssertionError, retryable: true) 
```

**Expected behavior**
Process should respect the class list I provide.

**Reason**
The problem is with the `Temporal\Internal\Assert::valuesSubclassOf` method. PHP's `instanceof` won't work here:

```php

$v = \InvalidArgumentException::class;
$of = \Throwable::class;

var_dump($v instanceof $of); // false, because $v is just a string
var_dump(is_subclass_of($v, $of)); // true, it checks hierarchy internally
```

I'll submit a PR to fix this shortly.

SerafimArts:
Fixed in #96

PR #96, opened 2021-04-05
Fix retryable exceptions class list in RetryOptions

kozlice:
## What was changed
Fixed the retryable exceptions class list incorrect behavior.

1. Closes issue: https://github.com/temporalio/sdk-php/issues/95

2. How was this tested: 
There is no test case for `withNonRetryableExceptions`, so manual testing only.

3. Any docs updates needed?
No, the example provided in docs is correct.

### DIFF
--- a/src/Common/RetryOptions.php
+++ b/src/Common/RetryOptions.php
@@ -199,7 +199,7 @@ class RetryOptions extends Options
     #[Pure]
     public function withNonRetryableExceptions(array $exceptions): self
     {
-        assert(Assert::valuesSubclassOf($exceptions, \Throwable::class));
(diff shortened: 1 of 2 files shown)


SerafimArts:
Yes, that should fix the described error. Thanks!

SerafimArts (approved):

kozlice:
@SerafimArts Done.

I also realized that my solution won't allow to use `Throwable::class` itself in the list. Replaced `is_subclass_of` with `is_a` to avoid it.

SerafimArts:
@kozlice yea, the subclass does not include its own declared class. I also overlooked this flaw.

Issue #98, opened 2021-04-13
[Bug] PHP 7.4 docblock anotations don't work (or documentation missing?)


Rastusik:
Hi, I'm trying to run temporal-sdk with php 7.4 (which is minimum supported version according to composer.json), but the SDK is unable to parse my workflows because I'm not using PHP 8 style attributes, but older PHP 7.4 docblock style. 

**To Reproduce**
1. try to work with this workflow interface:
```
use Temporal\Workflow\WorkflowInterface;
use Temporal\Workflow\WorkflowMethod;

/**
 * @WorkflowInterface()
 */
interface SomeWorkflow
{
    /**
     * @return void
     * @WorkflowMethod(name="fork")
     */
    public function something();
}
```
4. When the workflow client tries to work with the mentioned worklfow, it throws this error: `Workflow class SomeWorkflow or one of his parents (i.e. class, interface or trait) must contain #[Temporal\Workflow\WorkflowInterface] attribute`

**Expected behavior**
Well, the attribute/annotation should be found if minimum supported PHP version is 7.4

SerafimArts:
Hello!
1) Do you install a `doctrine/annotations` driver for annotation work?
2) Attributes work on any version of PHP.

Rastusik:
Hi, yes, I have `doctrine/annotations:^1.12` installed

SerafimArts:
Hmm, this is weird as this functionality has been tested by unit tests: https://github.com/temporalio/sdk-php/blob/master/tests/Unit/Declaration/DeclarationTestCase.php#L33

Rastusik:
from what I've been able to find out it looks like the Spiral Attribute reader doesn't even use doctrine annotation reader

Rastusik:
but the spiral FallbackAttributeReader is being used, and it looks lik it doesn't work properly

SerafimArts:
> from what I've been able to find out it looks like the Spiral Attribute reader doesn't even use doctrine annotation reader

Used to read annotations if a dependency is installed: https://github.com/temporalio/sdk-php/blob/master/src/WorkerFactory.php#L299-L306

> but the spiral FallbackAttributeReader is being used, and it looks lik it doesn't work properly...

This is a PHP 8 "emulator" that allows you to read attributes on any version of the PHP.

Rastusik:
btw I'm creating the Workflow client according to documentation: 
```
fn (): WorkflowClientInterface => WorkflowClient::create(
[...]

PR #100, opened 2021-04-13
Annotations fix in workflow client

SerafimArts:
Fixes https://github.com/temporalio/sdk-php/issues/98

### DIFF
--- a/src/Client/WorkflowClient.php
+++ b/src/Client/WorkflowClient.php
@@ -11,7 +11,12 @@ declare(strict_types=1);
 
 namespace Temporal\Client;
 
+use Doctrine\Common\Annotations\Reader;
+use Spiral\Attributes\AnnotationReader;
 use Spiral\Attributes\AttributeReader;
+use Spiral\Attributes\Composite\Composite;
+use Spiral\Attributes\Composite\SelectiveReader;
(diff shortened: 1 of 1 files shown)


src/Client/WorkflowClient.php
```diff
@@ -54,7 +59,19 @@ public function __construct(
...
         $this->converter = $converter ?? DataConverter::createDefault();
         $this->starter = new WorkflowStarter($serviceClient, $this->converter, $this->clientOptions);
-        $this->reader = new WorkflowReader(new AttributeReader());
+        $this->reader = new WorkflowReader($this->createReader());
+    }
+
+    /**
+     * @return ReaderInterface
+     */
+    private function createReader(): ReaderInterface
```
wolfy-j, line 68: We need this method in the Annotations package. This is the default configuration for now.

wolfy-j (approved):

src/Client/WorkflowClient.php
SerafimArts: I agree. But for now, either so, or take it out separately as done in the Cycle ORM: https://github.com/cycle/annotated/blob/master/src/ReaderFactory.php

SerafimArts:
CI errors are not related to PR.

Issue #110, opened 2021-05-18
[Question] Injecting custom reader into worker factory.


Zylius:
Hello,

Is there any way to inject a custom reader into WorkerFactory/Worker?

We're switching over to temporal and I'd like to register the old activities/workflows using our old logic for now and move over to Temporal attributes as new activites/workflows appear.

But it seems, cause of all the final classes the only way to do this would be to copy over the WorkerFactory completely.

Is this https://github.com/temporalio/sdk-php/issues/88 dealing with the same issue?

Thanks for help!

wolfy-j:
Can you elaborate? What are you trying to achieve? 

You have to manually register each activity, so you can pull it from any DI container.

Zylius:
Yes sure, here's an example:
``` php
#!/usr/bin/env php
<?php

use Temporal\WorkerFactory;

class SimpleWorker
{
    public function simpleActivityMethod(): string
    {
        return 'simple_string';
    }
}

$existingData = [
    SimpleWorker::class => [
        'methodNames' => [
            'simpleActivityMethod' => 'simpleMethod'
        ],
        'prefix' => 'simple_'
    ]
];

$factory = WorkerFactory::create();
$worker = $factory->newWorker();

// How do I register the activity, without PHP8 attributes, so that its info gets loaded from
// $existingData variable ? 
$worker->registerActivityImplementations(new SimpleWorker());

$factory->run();
```
And a more elaborate explanation:  lets say I'm adding an Activity object to a Worker using [registerActivityImplementations](https://github.com/temporalio/sdk-php/blob/master/src/Worker/Worker.php#L133). It reads the activity method groups, names and other info based on PHP 8 style attributes. Since the Worker is created using Worker Factory it has a fixed reader implementation (currently [SelectiveReader or AttributeReader](https://github.com/temporalio/sdk-php/blob/f53f1627f63fca73bd0e533831a8104bacd44309/src/DataConverter/JsonConverter.php#L214)).

Lets say I have the activity method names list and activity prefixes in some array parameter in my DI (not PHP8 style attributes). How do I force [ActivityReader](https://github.com/temporalio/sdk-php/blob/f53f1627f63fca73bd0e533831a8104bacd44309/src/Internal/Declaration/Reader/ActivityReader.php) to read it from the array parameter instead of the PHP8 attributes?

I was postulating that you could inject your own Reader instance into the WorkerFactory. But it's final and impossible. The only other two solutions I could think off would be:
[...]

Issue #88, opened 2021-03-04
Improve service container


SerafimArts:
Service container and worker factory classes:
- https://github.com/temporalio/sdk-php/blob/a5a212793f1cb5555d692fb134c01242f8e6111c/src/Internal/ServiceContainer.php#L125
- https://github.com/temporalio/sdk-php/blob/a5a212793f1cb5555d692fb134c01242f8e6111c/src/WorkerFactory.php#L286

Should be rewritten using a clearer and more user-friendly approach (global PSR-11?).

We should think about solving this problem, because a certain set of services is also used in the client.

wolfy-j:
Closing for now since the current structure is stable.

PR #115, opened 2021-07-02
Add support for custom reader in WorkerFactory.

Zylius:
## What was changed

Introduced an ability to pass a custom reader, just like with data converter.

## Why?

To be able to read activty/workflow config not only from annotations, but preset arrays etc.

## Checklist

1. Closes https://github.com/temporalio/sdk-php/issues/110

2. How was this tested: Unit tests.

3. Any docs updates needed? No. Its self explanatory and no BC.

### DIFF
--- a/src/WorkerFactory.php
+++ b/src/WorkerFactory.php
@@ -154,11 +154,16 @@ final class WorkerFactory implements WorkerFactoryInterface, LoopInterface
     /**
      * @param DataConverterInterface $dataConverter
      * @param RPCConnectionInterface $rpc
+     * @param ReaderInterface|null $reader
      */
-    public function __construct(DataConverterInterface $dataConverter, RPCConnectionInterface $rpc)
-    {
+    public function __construct(
+        DataConverterInterface $dataConverter,
+        RPCConnectionInterface $rpc,
+        ReaderInterface $reader = null
+    ) {
         $this->converter = $dataConverter;
         $this->rpc = $rpc;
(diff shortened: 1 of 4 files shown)


SerafimArts (approved):
LGTM.

Im only confused by the cases when we want to replace not only the reader, but also something else. It turns out that you need to add one more fourth argument to the constructor. And then the fifth. Etc.

wolfy-j:
I guess we need to add a container. :)

Zylius:
Yeah, I was wondering in the issue created, if this https://github.com/temporalio/sdk-php/issues/88 doesn't concern the same problem? 

Anyways, should I try to upgrade the container instead of this PR or is this okay for now?

SerafimArts:
> Yeah, I was wondering in the issue created, if this #88 doesn't concern the same problem?

Yep

> Anyways, should I try to upgrade the container instead of this PR or is this okay for now?

Nope. This is a global and rather serious task that requires a redesign of part of the architecture.

wolfy-j:
After a lot of thinking I tend not to merge this code into the master. 

The correct solution will be to add a proper DI Container for the configuration rather than exporting individual options. But this is a larger task which requires our team...

Can you send us a PR that makes WorkerFactory not final, so you can extend it in the meantime? You can then create a test that simply overwrites the reader via the redefined method. Would that work?

In a long term, we will make a proper IoC with PSR interface.

wolfy-j:
I apologise for such feedback, but it is necessary to keep the code consistency in the long term.

PR #118, opened 2021-07-12
Add support for custom reader in WorkerFactory.

Zylius:
## What was changed

Introduced an ability to extend worker factory.

Dunno if I should make other `create*()` methods protected as well, besides reader.

## Why?

So a custom reader could be used. To be able to read activty/workflow config not only from annotations, but preset arrays etc.

## Checklist

1. Closes https://github.com/temporalio/sdk-php/issues/110

2. How was this tested: Unit tests.

3. Any docs updates needed? No. Its self explanatory and no BC.

### DIFF
--- a/src/WorkerFactory.php
+++ b/src/WorkerFactory.php
@@ -67,7 +67,7 @@ use Temporal\Worker\WorkerOptions;
  *
  * </code>
  */
-final class WorkerFactory implements WorkerFactoryInterface, LoopInterface
+class WorkerFactory implements WorkerFactoryInterface, LoopInterface
 {
     use EventEmitterTrait;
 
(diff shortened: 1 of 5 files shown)


Zylius:
Related comment: https://github.com/temporalio/sdk-php/pull/115#issuecomment-874200948

Zylius:
Hello, is this still in consideration? This issue is kinda blocking our work with temporal so I was thinking if we should just use our own fork of the sdk.

SerafimArts (approved):
Of all the possible options, I think this is the simplest. 

Moreover, when we add a container, such an implementation may not even break anything... Probably... :)

wolfy-j (approved):

Zylius:
Thank you guys! :bow:

PR #125, opened 2021-11-09
Fix child workflow namespace propagation

### LINKED ISSUE #104: [Bug] Namespace not being propagated into ChildWorflowOptions, but probably should be (according to comments)
**Describe the bug**
I have a parent workflow which creates some child workflows. When using custom namespace (registered properly in the workflow client) it isn't being propagated to the child worflows (or child workflow options), but it seems like it should be propagated according to this comment: https://github.com/temporalio/sdk-php/blob/master/src/Workflow/ChildWorkflowOptions.php#L44

```
/**
     * Namespace of the child workflow.
     *
     * Optional: the current workflow (parent)'s namespace will be used if this
     * is not provided.
     */
    #[Marshal(name: 'Namespace')]
    public string $namespace = ClientOptions::DEFAULT_NAMESPACE;
```

It also might be the case that I'm doing something wrong in my parent worflow.

**To Reproduce**

My broken parent workflow looked like this:
```php
final class DefaultForkingWorkflow implements ForkingWorkflow
{
    public function fork(Event $event, HandlerDefinitions $definitions)
    {
        $promise = Workflow::async(static function () use ($event, $definitions) {
            $promises = [];

            foreach ($definitions as $definition) {
                $options = ChildWorkflowOptions::new()
                    ->withParentClosePolicy(ParentClosePolicy::POLICY_ABANDON)
                    ->withSearchAttributes([
                        SearchAttributes::ATTR_EVENT_ID => $event->getMetadata(Id::class)->toString(),
                        SearchAttributes::ATTR_HANDLER_ID => $definition->getServiceId(),
                        SearchAttributes::ATTR_HANDLER_METHOD => $definition->getHandlerMethod(),
                    ]);
                $handlerWorkflow = Workflow::newChildWorkflowStub(
                    HandlerWorkflow::class,
                    $options
                );

                $promises[] = Workflow::asyncDetached(
                    static fn () => yield $handlerWorkflow->handle($event, $definition)
                );
            }

            return yield Promise::all($promises);
        });

        try {
            yield $promise;
        } catch (TemporalException $e) {
            //$logger->error('child workflow failed');
            throw $e;
        }
    }
}
```

With this, temporal was logging this error: `Unable to schedule child execution across namespace default.` The error is correct because I changed the default namespace.

To make it work properly, I needed to propagate the namespace name from outside:

```php
final class DefaultForkingWorkflow implements ForkingWorkflow
{
    public function fork(Event $event, HandlerDefinitions $definitions, string $namespace)
    {
        $promise = Workflow::async(static function () use ($event, $definitions, $namespace) {
            $promises = [];

            foreach ($definitions as $definition) {
                $options = ChildWorkflowOptions::new()
                    ->withParentClosePolicy(ParentClosePolicy::POLICY_ABANDON)
                    ->withNamespace($namespace)
                    ->withSearchAttributes([
                        SearchAttributes::ATTR_EVENT_ID => $event->getMetadata(Id::class)->toString(),
                        SearchAttributes::ATTR_HANDLER_ID => $definition->getServiceId(),
                        SearchAttributes::ATTR_HANDLER_METHOD => $definition->getHandlerMethod(),
                    ]);
                $handlerWorkflow = Workflow::newChildWorkflowStub(
                    HandlerWorkflow::class,
                    $options
                );

                $promises[] = Workflow::asyncDetached(
                    static fn () => yield $handlerWorkflow->handle($event, $definition)
                );
            }

            return yield Promise::all($promises);
        });

        try {
            yield $promise;
        } catch (TemporalException $e) {
            //$logger->error('child workflow failed');
            throw $e;
        }
    }
}
```

**Expected behavior**
I think there should be no explicit namespace propagation (at least according to the docblock comment in the ChildWorkflowOptions class).

**Versions (please complete the following information where relevant):**
 - OS: Linux in Docker for mac
 - Temporal Version:  1.8.2
 - are you using Docker or Kubernetes or building Temporal from source? - nope, only when I need to dig for errors

seregazhuk:
Child workflow should inherit parent namespace if not provided.

Resolves #104

### DIFF
--- a/src/Internal/Workflow/WorkflowContext.php
+++ b/src/Internal/Workflow/WorkflowContext.php
@@ -293,7 +293,7 @@ class WorkflowContext implements WorkflowContextInterface
         string $type,
         ChildWorkflowOptions $options = null
     ): ChildWorkflowStubInterface {
-        $options ??= new ChildWorkflowOptions();
+        $options ??= (new ChildWorkflowOptions())->withNamespace($this->getInfo()->namespace);
 
         return new ChildWorkflowStub($this->services->marshaller, $type, $options);
     }
@@ -304,11 +304,12 @@ class WorkflowContext implements WorkflowContextInterface
     public function newChildWorkflowStub(string $class, ChildWorkflowOptions $options = null): object
     {
         $workflow = $this->services->workflowsReader->fromClass($class);
+        $options = $options ?? (new ChildWorkflowOptions())->withNamespace($this->getInfo()->namespace);
 
         return new ChildWorkflowProxy(
             $class,
             $workflow,
-            $options ?? new ChildWorkflowOptions(),
+            $options,
             $this
         );
     }


wolfy-j (approved):

wolfy-j:
Can you merge master to it?

seregazhuk:
@wolfy-j done.

wolfy-j:
Yay!

PR #145, opened 2022-03-03
Introduce workflow factory

seregazhuk:
The first idea was a factory to return an instance of `WorkflowPrototype`. But in this case any factory should have a dependency of `WorkflowReader`. I think that from the client point of view it is better to have just a plain class that returns just an object. Than, inside the `StartWorkflow` it will be read and converted to a `WorkflowPrototype`.

Api is the following:

```php
$workflowFactory = new class implements WorkflowFactoryInterface {
    public function create(string $workflowName): ?object
    {
        if ($workflowName !== 'my-workflow') {
            return null;
        }

        return new
        /**
         * Support for PHP7.4
         * @Workflow\WorkflowInterface
         */
        #[Workflow\WorkflowInterface]
        class {
            /**
             * Support for PHP7.4
             * @Workflow\WorkflowMethod(name="SimpleWorkflow")
             */
            #[WorkflowMethod(name: 'SimpleWorkflow')]
            public function handler()
            {
                return 'hello';
            }
        };
    }
};

$worker = $this->factory->newWorker(
    'default',
    null,
    null,
    $workflowFactory
);
```

### DIFF
--- a/src/Internal/Transport/Router/StartWorkflow.php
+++ b/src/Internal/Transport/Router/StartWorkflow.php
@@ -83,10 +83,7 @@ final class StartWorkflow extends Route
      */
     private function findWorkflowOrFail(WorkflowInfo $info): WorkflowPrototype
     {
-        $workflow = $this->services->workflows->find($info->type->name);
-        if ($workflow === null) {
-            $workflow = $this->workflowFactory->create($info->type->name);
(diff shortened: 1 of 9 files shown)


seregazhuk:
Closing because it requires changes on Go side.

PR #161, opened 2022-04-19
Support for full set of promise functions

OlegKuro:
## What was changed
I've added support for the rest of available [traditional promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#static_methods) methods 

## Why?
I feel uncomfortable about using `Promise::all([])`, when I need a fulfilled promise 😓 
So I decided to add `Promise::resolve` and the rest of forgotten methods

### DIFF
--- a/src/Promise.php
+++ b/src/Promise.php
@@ -18,6 +18,9 @@ use function React\Promise\any;
 use function React\Promise\map;
 use function React\Promise\reduce;
 use function React\Promise\some;
+use function React\Promise\resolve;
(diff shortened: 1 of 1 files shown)


OlegKuro:
@seregazhuk Hi!
Could you please take a look at this PR? 🥺

seregazhuk:
For me it is OK, ping @wolfy-j

wolfy-j (approved):

PR #164, opened 2022-04-26
Use "proto" for RR_CODEC

seregazhuk:


src/WorkerFactory.php
```diff
@@ -363,8 +363,8 @@ private function createMarshaller(ReaderInterface $reader): MarshallerInterface
      */
     private function createCodec(): CodecInterface
     {
-        // todo: make it better
         switch ($_SERVER['RR_CODEC'] ?? null) {
```
rustatian, line 366: We don't have a JsonCodec anymore in the RR, only proto. So, I guess the JsonCodec can be safely removed as we removed it from the golang since version 1.1. Previously RR used it only for the debugging.

seregazhuk: Yes, but currently it is used in tests - https://github.com/temporalio/sdk-php/blob/master/tests/Fixtures/WorkerMock.php#L43

So, we still keep it.

wolfy-j:
Tests :(

seregazhuk:
@wolfy-j fixed tests

rustatian (approved):

PR #181, opened 2022-05-30
[FIX] Correct encoding of null values, strip names of arguments

### LINKED ISSUE #180: [Bug] Small bugs in EncodedValues

I've build some workflows test based on your Test Framework, and noticed two small bugs:

- wrong decoding of `null` values
- wrong encoding of named arguments, i.e.
```
yield $this->activity->method(arg1: 'foo', arg2: 'bar');
```

I'll create PR with fixes.

mzavatsky:
## What was changed
Small fixes in Encoded values:
1. getValue fails if i.e. `$this->values[0] === null` 
2. fromValues fails when we call i.e. activity with named arguments, `$ev->values` will have named keys

1. Closes #180

### DIFF
--- a/src/DataConverter/EncodedValues.php
+++ b/src/DataConverter/EncodedValues.php
@@ -66,7 +66,7 @@ class EncodedValues implements ValuesInterface
      */
     public function getValue(int $index, $type = null)
     {
-        if (isset($this->values[$index])) {
+        if (is_array($this->values) && array_key_exists($index, $this->values)) {
             return $this->values[$index];
         }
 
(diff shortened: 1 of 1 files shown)


seregazhuk (approved):
Thank you! 👍

rustatian (approved):

seregazhuk:
@mzavatsky can you please apply cla check with this link https://cla-assistant.io/temporalio/sdk-php?pullRequest=181

mzavatsky:
> @mzavatsky can you please apply cla check with this link https://cla-assistant.io/temporalio/sdk-php?pullRequest=181

Done! I've checked it before, but the problem was in e-mail address, now everything is fine!

PR #201, opened 2022-06-24
Enums php81

cv65kr:
## What was changed
Add new marshal type `EnumType` which is available for PHP 8.1

## Why?
To use all functionality provided by new version of PHP

## Checklist
1. How was this tested:
Unit + Manual


seregazhuk:
@cv65kr thank you! 🔥 can you check tests, please?

cv65kr:
@seregazhuk I suppressed warning for return type because psalm is more restrictive then native PHP. Please rerun tests, I not tested it because I don't have PHP 7.4 on my local machine

seregazhuk (approved):

PR #209, opened 2022-07-21
Fix null pointer

cv65kr:
## What was changed
In 399 line null pointer can occur.
```
if ($response->getHistory()->getEvents()->count() === 0) {
```

getHistory(): \Temporal\Api\History\V1\History|null

## Why?
image

Bug

## Checklist
1. How was this tested:
In my project I have a multiple integration tests, so errors was visible.

### DIFF
--- a/src/Internal/Client/WorkflowStub.php
+++ b/src/Internal/Client/WorkflowStub.php
@@ -392,6 +392,10 @@ final class WorkflowStub implements WorkflowStubInterface
                 }
             }
 
+            if ($response->getHistory() === null) {
+                continue;
+            }
+
             if ($response->getHistory()->getEvents()->count() === 0) {
                 continue;
             }


seregazhuk:
@cv65kr can you add a test for it? If there are no details in response may be it is not safe just to ignore it? 🤔

cv65kr:
@seregazhuk It's extremely hard to add test for that. My assumptions is that not generate any BC break.

Generally next line `$response->getHistory()->getEvents()->count() === 0` means that if no events in the history we should continue this infinity loop.
And my fix doing exactly the same, because history is empty so events not exists yet.

seregazhuk:
@cv65kr I have created a PR with your changes but added a test there. Closing this one. If you have any suggestions - welcome 😉 
https://github.com/temporalio/sdk-php/pull/215

PR #210, opened 2022-07-23
Do not pass nullable result to marshaller

kafkiansky:
## What was changed
Changed behavior for data marshalling.  

Now marshaller throws the `TypeError` when we trying to marshal nullable structure:

```php
final class DeclineReason
{
     public function __construct(
        public readonly string $code,
        public readonly ?string $description = null,
     ) {}  
}

final class Transaction
{
    public function __construct(public readonly ?DeclineReason $reason = null)
    {
    }
}
```

### DIFF
--- a/src/Internal/Marshaller/Mapper/AttributeMapper.php
+++ b/src/Internal/Marshaller/Mapper/AttributeMapper.php
@@ -170,7 +170,7 @@ class AttributeMapper implements MapperInterface
                 return null;
             }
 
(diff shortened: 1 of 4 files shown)


wolfy-j:
Would you be able to add a test as well?

kafkiansky:
> Would you be able to add a test as well?

Done.

seregazhuk (approved):

seregazhuk:
@kafkiansky  thank you! Merged 👍

PR #214, opened 2022-07-27
Use RR kv only for testing

seregazhuk:


wolfy-j:
Need a demo first.

seregazhuk:
@wolfy-j is described in [readme](https://github.com/temporalio/sdk-php/blob/1331889106436e57ec4232460604c9fc504b6b51/testing/Readme.md#mocking-activities). With a simple interface just to mock completion or failure. I see it as a proof of concept. Later can be approved with a real mocking interface. Like expect this activity to be invoked N times and return/fail with something.

rustatian (approved):

PR #217, opened 2022-07-29
Possibility to configure temporal test server host

cv65kr:
## Why?
To have possibility to change default port for temporal test server.
Also:
Since output of RR is not visible in PHPunit sometimes it's hard to figure out where is the error. So you can switch test server to real temporal server and see in UI what's happened.
Alternatively Datadog offers https://github.com/temporalio/temporalite own server which also can be use for tests purposes.

### DIFF
--- a/testing/src/Environment.php
+++ b/testing/src/Environment.php
@@ -42,9 +42,11 @@ final class Environment
             $this->output->writeln('<info>done.</info>');
         }
 
+        $temporalAddress = getenv('TEMPORAL_ADDRESS') ?: '127.0.0.1:7233';
+
         $this->output->write('Starting Temporal test server... ');
         $this->temporalServerProcess = new Process(
-            [$this->systemInfo->temporalServerExecutable, getenv('TEMPORAL_ADDRESS'), '--enable-time-skipping']
+            [$this->systemInfo->temporalServerExecutable, $temporalAddress, '--enable-time-skipping']
         );
         $this->temporalServerProcess->setTimeout(10);
         $this->temporalServerProcess->start();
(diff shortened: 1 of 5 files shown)


seregazhuk:
Maybe it can be moved to env variable in `phpunit.xml` file? 🤔 Instead of hardcoding or passing around variables.

cv65kr:
That was my original idea, but I was a bit confused why it was hard coded on beginning 🤔 Anyway updated

testing/src/Environment.php
```diff
@@ -44,7 +44,7 @@ public function start(string $rrCommand = null): void
 
         $this->output->write('Starting Temporal test server... ');
         $this->temporalServerProcess = new Process(
-            [$this->systemInfo->temporalServerExecutable, 7233, '--enable-time-skipping']
+            [$this->systemInfo->temporalServerExecutable, getenv('TEMPORAL_ADDRESS'), '--enable-time-skipping']
```
seregazhuk, line 47: Maybe we can add a default value here? Like this:

```php
getenv('TEMPORAL_ADDRESS') ?: '127.0.0.1:7233`
```

Or pass the variable into `Environment::create()` constructor and hide this "default logic" there. The same about trait and `WorkflowTestCase`. Let's add deafult value there.

seregazhuk (approved):

testing/src/Environment.php
cv65kr: I added default value :)

cv65kr:
@seregazhuk I think I found issue with CI, let's try with CI

seregazhuk (approved):
@cv65kr thank you! It passes all checks 💪 merging 🚀

Issue #203, opened 2022-07-08
[Bug] Workflow shouldn't wait for activities unless it is explicitly said with yield


seregazhuk:
Having such workflow, that calls activity and doesn't use (not waits) for its result (no `yield` keyword):

```php
#[WorkflowMethod]
public function start()
{
    $activity = Workflow::newActivityStub(
        SleepActivity::class,
        ActivityOptions::new()->withStartToCloseTimeout(20)
    );
    $activity->wait();

    return 'Done';
}
```

Inside activity we have a simple sleep():

```php
#[\Temporal\Activity\ActivityMethod]
public function wait(): void
{
    sleep(5);
}
```

Expected behavior: workflow finishes and doesn't wait for 5 seconds.
Actual behavior: workflow waits for 5 seconds.

PR #218, opened 2022-08-01
fix: don't wait implicitly for child scope

### LINKED ISSUE #207: [Bug] Order of declaration async tasks completely change the execution flow of Workflow

I was trying to create a Workflow that could be stopped by sending the signal and that simultaneously run several tasks.
The idea behind simultaneous tasks consists of waiting for a signal with data and trying to get this data in case that signal wasn't called.
So generally workflow consists of 4 asynchronous tasks:

1. "main" - a task that contains all other code and can be rejected by sending a specific signal
2. Rejection task - checks for signal and reject "main" task when the signal is received
3. Task waiting for the signal with required data
4. Task that calls activity to get the same data as the previous task. Also contains a timer and infinitely calls the activity until get the required data.

### Describe the bug

Depending on the order of declaring the two last tasks (that wait for a signal with data and that call activity to get that data) "main" task can either be rejected when one of the tasks finishes or can proceed successfully.

### Minimal Reproduction

```php
    private bool $isSignaled = false;
    private bool $isRejected = false;

    private Workflow\CancellationScopeInterface $first;
    private Workflow\CancellationScopeInterface $second;

    #[WorkflowMethod]
    public function execute(): \Generator
    {
        $workflowPromise = Workflow::async(
            function () {
                $this->first = Workflow::async(
                    function () {
                        yield Workflow::await(fn() => $this->isSignaled);

                        $this->second->cancel();
                        print "First finished";
                    }
                );

                $this->second = Workflow::async(
                    function () {
                        yield Workflow::timer(5);

                        $this->first->cancel();
                        print "Second finished";
                    }
                );

                yield Promise::race([$this->second, $this->first]);

                print "Main finished";
            }
        )->then(onRejected: fn() => print "Main rejected!");

        Workflow::async(
            function () use ($workflowPromise) {
                yield Workflow::await(fn() => $this->isRejected);
                $workflowPromise->cancel();
            }
        );

        yield $workflowPromise;
    }

    #[Workflow\SignalMethod]
    public function signal(): void
    {
        $this->isSignaled = true;
    }
```

In such flow, the output is following. When the second task finishes after the timer the message "Main rejected!" is displayed.
In case I send a signal before the timer is finished the message "Main finished" is displayed which is the expected result.

But when I declare $this->second before $this->first in both cases "Main finished" is displayed.

Expected behavior: order of declaration async tasks doesn't affect the normal flow of execution.
Actual behavior: in some order of declaration async tasks "main" task is rejected.

### Additional context
image
image

seregazhuk:
Resolves #207, #203


wolfy-j:
Wonder what performance benefits we also going to get. We have to prep users for this release. This is major change.

wolfy-j (approved):

seregazhuk:
>Wonder what performance benefits we also going to get. 

@wolfy-j any ideas on how to "measure" it?

wolfy-j:
It's already pretty fast to measure on a single workflow, we can try to measure the duration of `WorkflowTask` on a workflow with many async tasks (ideally coroutines). Not sure.

PR #219, opened 2022-08-02
Stop tests and display error if RoadRunner start failed

shanginn:
## What was changed

Added additional check for RR in the Temporal\Testing\Environment::start method

## Why?

RoadRunner could fail to start for many reasons, for example:
- rr binary doesn't exist (but in that case we could download it I suppose)
- config file not found
- error in the worker setup

It is useful to stop the tests right away and display the error, otherwise the tests are just hang
## Checklist

1. Closes 

2. How was this tested:
- Delete rr binary
- Set wrong path to config in rrCommand (`$environment->start('./rr serve -c missing.rr.yaml');`)
- Put die, or some other error in the worker

3. Any docs updates needed?

### DIFF
--- a/.github/workflows/code-style.yml
+++ b/.github/workflows/code-style.yml
@@ -25,20 +25,18 @@ jobs:
           fetch-depth: 1
 
       - name: Get Composer Cache Directory
-        # Docs: <https://github.com/actions/cache/blob/master/examples.md#php---composer>
         id: composer-cache
         run: echo "::set-output name=dir::$(composer config cache-files-dir)"
 
       - name: Cache Dependencies
-        # Docs: <https://github.com/actions/cache/blob/master/examples.md#php---composer>
-        uses: actions/cache@v1
+        uses: actions/cache@v3
         with:
           path: ${{ steps.composer-cache.outputs.dir }}
-          key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
-          restore-keys: ${{ runner.os }}-composer-
+          key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
+          restore-keys: php-${{ matrix.php }}-${{ runner.os }}-composer-
 
       - name: Install Composer Dependencies
-        run: composer install --prefer-dist --no-interaction --no-suggest
+        run: composer install --prefer-dist --no-interaction
 
       - name: Run Tests
         run: vendor/bin/psalm
--- a/.github/workflows/run-test-suite.yml
+++ b/.github/workflows/run-test-suite.yml
@@ -5,13 +5,17 @@ on:
         required: true
         type: string
       fail-fast:
-        required: true
+        required: false
         type: boolean
         default: true
       test-timeout:
         required: false
         type: number
         default: 10
+      run-temporal-test-server:
+        required: false
+        type: boolean
+        default: true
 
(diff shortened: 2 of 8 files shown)


seregazhuk:
@shanginn thank you! 👍 
Can you please follow this link and sign CLA - https://github.com/temporalio/sdk-php/pull/219#issuecomment-1202065606

seregazhuk:
Looks like some tests are failing.

shanginn:
> Looks like some tests are failing.

looks like some of them are failing because rr is unable to start)
I will take a look at it tomorrow

shanginn:
Forgot about Feature test, added it too. Let's try again :)

seregazhuk:
@shanginn don't see a feature test 🙈 Forgot to push it?

shanginn:
what I meant is that I forgot to add "Download Roadrunner" step to the "Feature tests" CI part.

Do you think I should write a test for it? I mean it is clearly failing without rr 😅

I was thinking about feature test for it and haven't come up with a good idea how I could do it, do you have some thoughts?

But the main question is why it is failing on some test cases, I will try to debug it

shanginn:
ok, so I learn a few things: it's failing on `--prefer-lowest` cases, because rr downloads beta version and it seems to not downloading correctly. But `--prefer-lowest` is not a stability option, but it is used as one in your matrix config and (I think) roadrunner also uses it as stability marker, but it's wrong.

> Even [composer docs](https://getcomposer.org/doc/03-cli.md) hints at using `--prefer-lowest` alongside `--prefer-stable` option. Also in your `composer.json` `prefer-stable` is set to true, so it is always true regardless the `stability` matrix option.

But maybe rr failing to detect correct version because v2.0.0 (which is downloaded because of --prefer-lowest) does not have `--stability` option yet. And I guess upping rr version in the composer.json is not an option for such a change.

But even bumping the version to 2.1/2.2/2.3 doesn't help. the latest stable 2.* version is downloaded, but rr is failing to install.

It is not really a CI PR, maybe you have some suggestions? Anyways I will take another shot at it tomorrow

shanginn:
ugh :|
I don't know why it's failing after successfully downloading rr binary

seregazhuk:
@shanginn thank you for your work! 🔥  
Maybe there is no need to run RR/server in unit tests? 🤔 For example, we can define environment `RUN_TEMPORAL_TEST_SERVER` variable and check it `bootstrap.php`:

```php
if (getenv('RUN_TEMPORAL_TEST_SERVER') !== false) {
    $environment = Environment::create();
    $environment->start('./rr serve -c .rr.silent.yaml -w tests');
    register_shutdown_function(fn() => $environment->stop());
}
```

What do you think about it?

shanginn:
yeah, good idea! how about this?

it is still failing sometime on `rr` execution, but I have no idea why.

also you may think this is a strange way to implement this: I added 2 steps **with** and **without** env. just passing env every time is not an option, because env `"false"` is a string and `getenv` is true. so it's either this or some ternary env before phpunit command which is neither readable nor great.

seregazhuk:
Looks great! 💪 Can you update testing readme with a new env variable, please?

shanginn:
yep, sure. I think it should not be the default option for people, so I added a separate entry about it

seregazhuk (approved):
Merging 👍

Issue #211, opened 2022-07-25
[Feature Request] Ability to use readonly properties in workflow and activity inputs.


kafkiansky:
Now each property is set individually after the object has been created. This does not work with `readonly` fields.

shanginn:
I wanted to fix this, but was surprised, because readonly properties in DTO's are actually work.

I created demonstration [here](https://github.com/shanginn/temporal-php-worker/tree/feature-sdk-211-issue-demostration) for workflow argument and activity argument.

And pull request with unit test [here](https://github.com/temporalio/sdk-php/pull/232)

PR #232, opened 2022-08-09
Issue 211. Test for readonly properties

shanginn:
Ok, I remember that readonly didn't work for me, but now they are working, so I created a test case for it. In order for it to work we need to add php 8.1 in CI matrix settings.

But php 8.1 will not work with current lower versions.

The questions is - do you want to add php 8.1 support? If no, I can just add the test and it will be waiting for the version upgrade.

There is another question: how should I go about testing this behavior: with **marshaller** or with **payload converter**? I created both, but I think that one of them is enough. Or both is good?

## What was changed
Added test that covers [readonly properties issue](https://github.com/temporalio/sdk-php/issues/211)

## Why?
Because issue is open and I though readonly properties would not work, but surprisingly they are working just fine

## Checklist

1. Closes [issue 221](https://github.com/temporalio/sdk-php/issues/211)

2. How was this tested:
With unit test

### DIFF
--- a/.github/workflows/run-test-suite.yml
+++ b/.github/workflows/run-test-suite.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       fail-fast: ${{ inputs.fail-fast }}
       matrix:
-        php: [ 7.4, 8.0 ]
+        php: [ 7.4, 8.0, 8.1 ]
         os: [ ubuntu-latest , windows-latest]
         dependencies: [ lowest , highest ]
     steps:
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "ext-curl": "*",
         "ext-json": "*",
         "google/common-protos": "^1.3",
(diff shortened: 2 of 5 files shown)


seregazhuk:
@shanginn have you tried functional tests with your changes? It fails, I think there are errors when running worker under RR.
https://github.com/temporalio/sdk-php/runs/7752350196?check_suite_focus=true

shanginn:
I did check it under my branch, and it [went fine](https://github.com/shanginn/sdk-php/runs/7747314773?check_suite_focus=true)

shanginn:
I rebased (or do you prefer merge?) this branch to the current master. in my fork [all the test are good](https://github.com/shanginn/sdk-php/actions/runs/2836983394)

shanginn:
😢

seregazhuk (approved):
Thank you! 🔥

Issue #112, opened 2021-05-26
[Question] Arrays not allowed as a return type?


Zylius:
Hello once again, I wanted to ask about [this](https://github.com/temporalio/sdk-php/blob/f53f1627f63fca73bd0e533831a8104bacd44309/src/DataConverter/Type.php#L104) line.

I was curious why is that. I'm sure there's a specific reason for it. I could still wrap it in a simple container object.
An example:

```php
#[ActivityInterface]
class WorkflowConfigActivity
{
    #[ActivityMethod]
    public function getTransactionConfigs(): array // Doesn't work
    {
        return ['my_config' => 'best_config'];
    }
}
```

SerafimArts:
@Zylius hello! The reason is simple: Any iterator is generic, i.e. a composite type.

It is impossible to write in PHP:
```php
public function getTransactionConfigs(): array<string, string>
{
    // ...
}
```

In your case, there are two restrictions at the same time:
1) The absence of the concept of a hash map in JSON (There are only objects)
2) Inability to deduce an unambiguous type from return typehint.

In summary: Recommended to use DTOs to transfer composite objects between code.
Similar question: https://github.com/temporalio/sdk-php/issues/93

Zylius:
Ahh gotya! Closing this ;) Thanks.

Issue #93, opened 2021-03-22
[Bug] Activity return an associative array, but in workflow, the return value is parsed as an object


khoerintus:
**Describe the bug**
Activity return associative array, but when the return result is used in workflow, it became an object, not an array.

**To Reproduce**
1. In Activity method, return associative array. Example:
```
        return [
            'Peter' => '123',
            'Ben' => '234',
            'Joe' => '345'
        ];
```
2. Use the return value as associative array in workflow. Example:
```
        $simple = Workflow::newActivityStub(
            SimpleActivity::class,
            ActivityOptions::new()
                ->withStartToCloseTimeout(300)
                ->withRetryOptions(
                    RetryOptions::new()->withMaximumAttempts(1)
                )
        );
        $assocArray = yield $simple->getAssocArray($input);

        return $assocArray['Peter'];
```

**Expected behavior**
I can use the return value, as an array, not as an object.

**Screenshots/Terminal ouput**
```
WARN    temporal        Failed to process workflow task.        {"Namespace": "default", "TaskQueue": "default", "WorkerID": "209439@khoerintus-Lenovo-G40-80@default@1", "WorkflowType": "SimpleWorkflow.handle", "WorkflowID": "d411c5ae-32a7-4879-873b-704dc270eef8", "RunID": "dcf2fa14-ceb2-4c99-9b17-a7d214e4d84f", "Attempt": 9, "Error": "Cannot use object of type stdClass as array (type: Error, retryable: true)"}

[...]

PR #234, opened 2022-08-11
Ability to return associative array from an activity

### LINKED ISSUE #117: [Bug] Wrong activity response type
**Describe the bug**
I send an array from my activity as a result, but in promise step **then** I get **stdClass** instead of array.

**To Reproduce**
1. Create activity that will send an associative array.
2. Create workflow where you call activity and add post-processing in **then** step.
3. Try to get value from **result** as if it array.

**Expected behavior**
I get value

**Expected behavior**
I get an error `PanicError: Cannot use object of type stdClass as array (type: Error, retryable: true)`

**Screenshots/Terminal ouput**
image

You can use this snippet:
```php
#[\Temporal\Activity\ActivityInterface(prefix: "app.")]
class CommonActivity
{
    public function slow(string $name): array
    {
        return [
            'result' => 'Hello, ' . $name,
        ];
    }
}
```

```php
#[\Temporal\Workflow\WorkflowInterface]
final class FastWorkflow
{
    #[\Temporal\Workflow\WorkflowMethod("wf")]
    public function run(string $name, int $count): \Generator
    {
        $activity = Workflow::newActivityStub(
            CommonActivity::class
        );

        $activity->slow($name)->then(function ($result) {
            $result['result'];
        });

        return 0;
    }
}
```

As I understand the problem is locating [here](https://github.com/temporalio/sdk-php/blob/master/src/DataConverter/JsonConverter.php#L86). But I don't know how to solve it.

shanginn:
## What was changed

Added ability to return associative array from activities with JsonEncoder.
## Why?

Although @SerafimArts explains in every related issue that json has no support for hashmaps or associative arrays (and this is indisputable true), but we in PHP do have support for it.

And in case of PHP workflow, PHP activity and PHP sdk we control input, output and the type, so I say we should be able to use associative arrays.

Yes, it is json, but we will use stock json_decode to get the required associative array.

Otherwise the SDK should forbid typehint activities with the `array` type. It is counterintuitive: if the `newActivityStub` returns `T` and I have an `array` declared as a return type for a function of `T` I would like to get it, or get an error when I'm trying to create proxy of `T`

## Checklist

1. Closes
#117
#93
#112

2. How was this tested:

With unit test and in [my worker](https://github.com/shanginn/temporal-php-worker/tree/feature-test-assoc-arrays)
3. Any docs updates needed?

I read in these issues that it is documented behavior. I guess this part should be updated is this PR is accepted

### DIFF
--- a/src/DataConverter/JsonConverter.php
+++ b/src/DataConverter/JsonConverter.php
@@ -83,7 +83,12 @@ class JsonConverter extends Converter
     public function fromPayload(Payload $payload, Type $type)
     {
         try {
-            $data = \json_decode($payload->getData(), false, 512, self::JSON_FLAGS);
+            $data = \json_decode(
+                $payload->getData(),
+                $type->getName() === Type::TYPE_ARRAY,
+                512,
+                self::JSON_FLAGS
+            );
         } catch (\Throwable $e) {
             throw new DataConverterException($e->getMessage(), $e->getCode(), $e);
         }
--- a/src/DataConverter/Type.php
+++ b/src/DataConverter/Type.php
@@ -107,7 +107,7 @@ final class Type
             }
 
             // Traversable types (i.e. Generator) not allowed
-            if (!$name instanceof \Traversable && $name !== 'array' && $name !== 'iterable') {
+            if (!$name instanceof \Traversable && $name !== 'iterable') {
                 return new self($type->getName(), $type->allowsNull());
             }
         }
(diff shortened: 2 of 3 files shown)


seregazhuk (approved):
Thank you! Will be released in v2.0 because it is actually a breaking change.

shanginn:
nice! thanks!
looking forward to use it :)

PR #235, opened 2022-08-22
Add error output for temporal test server

dehbka:
## Why?
Got error on temporal test server starting but not handle it
image


seregazhuk (approved):
Thank you! 🔥

PR #241, opened 2022-09-02
[Testing] Allow mocking activity result with complex objects

cappuc:
## What was changed
This changes improve activity mocking with support for complex return types (and not just native types)
The changed made to allow this are:
- Added an `ActivityInvocationResult` class which encode the response value with the `DataConverter` and wrap it in `EncodedValues` (like a real response) and then save the `EncodedValues` payload to the cache.
- In `ActivityInvocationCacheInterface@execute` method check for the new `ActivityInvocationResult` object and return the encoded values a response

The other changes are only required for passing the `DataConverter` to the cache and to allow an easy override of the cache during test worker construction

## Why?
Right now activity mocking works only with native types and not with more complex object.

## Checklist
How was this tested:
- All existing tests pass
- The added parameters are all optional so there shouldn't be any breaking change

### DIFF
--- a/src/Worker/ActivityInvocationCache/ActivityInvocationResult.php
+++ b/src/Worker/ActivityInvocationCache/ActivityInvocationResult.php
@@ -1,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Temporal\Worker\ActivityInvocationCache;
+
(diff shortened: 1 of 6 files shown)


seregazhuk (approved):
Thank you! 🔥

PR #245, opened 2022-09-23
Support nullable array types, `?array`

roquie:
## What was changed

When developer using class as DTO with public properties to pass values to workflow with property type `?array` it got an error:

```
InvalidArgumentException: Passed value must be a type of array, but null given in /backend/vendor/temporal/sdk/src/Internal/Marshaller/Type/ArrayType.php:64\nStack trace:\n#0 /backend/vendor/temporal/sdk/src/Internal/Marshaller/Mapper/AttributeMapper.php(195): Temporal\\Internal\\Marshaller\\Type\\ArrayType->parse(NULL, NULL)
```

This error has been fixed by this PR.


seregazhuk:
@roquie thank you for PR 💪 Maybe you can add a simple test for it?

rustatian (approved):

roxblnfk (approved):

PR #250, opened 2022-09-29
Send signal to workflow via parent interface

smelesh:
## What was changed

Add an ability to create workflow stub without a handler method, so that signals can be sent to a running workflow via parent reusable interface.

## Why?

[Workflow interface inheritance](https://docs.temporal.io/php/workflows/#workflow-interface-inheritance) is broken because workflow stub can not be created without a handler method annotated with `WorkflowMethod`.

Handler method is needed only to start a new workflow, but it's okay to send signals or queries to other workflows by stub created from a reusable parent interface.

## Checklist

1. How was this tested:

Integration tests


rustatian:
Friendly ping @wolfy-j / @roxblnfk

roxblnfk (approved):
I'm sorry for the delay. Looks very good.

PR #259, opened 2022-10-18
feat: allow to return array of objects

seregazhuk:
## What was changed
Allowed possibility to return a list of object from activity/workflow. 
When executing workflow:
```php
/** @var Message[] $result */
$result = $client->start($workflow, 'John')->getResult(Type::arrayOf(Message::class));
```

When executing activity within a workflow:
```php
/** @var Message[] $result */
$result = yield $activity->execute('myActivity', [$input], Type::arrayOf(Message::class));
```

## Why?
Now, we cant return a list of objects. It will always be an array of arrays.

### How was this tested:
With a functional test.

### Any docs updates needed?
Not sure that it is covered in the docs 🤔

### DIFF
--- a/src/DataConverter/JsonConverter.php
+++ b/src/DataConverter/JsonConverter.php
@@ -150,13 +150,15 @@ class JsonConverter extends Converter
             }
 
             if ($type->isArrayOf()) {
-                return array_map(
-                    fn($rawObject) => $this->unmarshal($rawObject, $reflection->newInstanceWithoutConstructor()),
-                    $data
-                );
+                $result = [];
+                foreach ($data as $key => $value) {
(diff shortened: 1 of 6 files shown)


wolfy-j (approved):

rustatian:
@seregazhuk Please, attach a PR to the docs (or samples-php). Converting to draft until that, thanks.

Neznaykas:
will this solution be implemented?

I also faced the problem of newInstanceWithoutConstructor when using sub child of child workflow

roxblnfk:
> will this solution be implemented?

Yes. A bit later 👌

roxblnfk (approved):

roxblnfk:
Good PR, thanks.

PR #263, opened 2022-11-14
Wrapper for multiple task queues

cv65kr:
## What was changed
- Possibility to use multiple roadrunners instances, because we can have separate worker per task queue
- Possibility to add environment variables for RR process

## Why?
For more flexible testing

## Checklist
1. Any docs updates needed?
Yes


testing/src/Environment.php
```diff
@@ -66,10 +66,11 @@ public function startTemporalTestServer(int $commandTimeout = 10): void
         }
     }
 
-    public function startRoadRunner(string $rrCommand = null, int $commandTimeout = 10): void
+    public function startRoadRunner(string $rrCommand = null, int $commandTimeout = 10, array $envs = []): void
     {
         $this->roadRunnerProcess = new Process(
-            $rrCommand ? explode(' ', $rrCommand) : [$this->systemInfo->rrExecutable, 'serve']
+            command: $rrCommand ? explode(' ', $rrCommand) : [$this->systemInfo->rrExecutable, 'serve'],
+            env: $envs
```
roxblnfk, line 73: > \* \@param array|null $env The environment variables or null to use the same environment as the current PHP process
> *\Symfony\Component\Process\Process::__construct()*

Are you sure the default value should be an empty array?
I think it will change the current behavior.

roxblnfk (commented):
Hello.
Thank you for the PR.

testing/src/Environment.php
cv65kr: It not change the behaviour - https://github.com/symfony/process/blob/6.2/Process.php#L160

cv65kr:
@roxblnfk Forgive me for being late with changes, big thanks for review. PR updated 👍

wolfy-j (approved):

testing/src/Environments.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Temporal\Testing;
+
+final class Environments
```
roxblnfk, line 7: Hi @cv65kr.
I am not sure that class is needed.
We can replace it with closure that will stop all Environment objects.
What do you think?

wolfy-j:
@cv65kr  Is this PR still relevant? Can you please help with @roxblnfk questions?

cv65kr:
@wolfy-j Will look in this week

testing/src/Environments.php
kajetan-koltuniak-tg: @roxblnfk fixed

cv65kr: @roxblnfk fixed

roxblnfk (approved):

testing/Readme.md
```diff
@@ -47,7 +62,6 @@ $environment = new Environment(
     ),
     SystemInfo::detect(),
 );
-```
```
roxblnfk, line 50: Lost end of code block

PR #265, opened 2022-11-25
Support newer `google/common-protos` package versions 

cappuc:
## What was changed

- Extended `google/common-protos` version range to add support for v2 and v3
- Updated version ranges of some dev dependencies to match version ranges of dependencies

## Why?
- Support newer `google/common-protos` versions
- Old dev dependencies versions prevents the installation of newer dependencies versions (in supported ranges) so the newer versions are not tested


composer.json
```diff
@@ -54,13 +54,13 @@
         "dereuromark/composer-prefer-lowest": "^0.1.10",
         "doctrine/annotations": "^1.11",
         "friendsofphp/php-cs-fixer": "^3.0",
-        "illuminate/support": "^7.0",
+        "illuminate/support": "^7.0 || ^8.0 || ^9.0",
```
roxblnfk, line 57:
```suggestion
        "illuminate/support": "^9.0",
```

composer.json
```diff
@@ -54,13 +54,13 @@
...
         "jetbrains/phpstorm-attributes": "dev-master@dev",
         "laminas/laminas-code": "^4.0",
         "monolog/monolog": "^2.1 || ^3.0",
         "phpunit/phpunit": "^9.5.21",
-        "symfony/translation": "5.4.*",
-        "symfony/var-dumper": "^5.1",
+        "symfony/translation": "^4.4.20 || ^5.0 || ^6.0",
+        "symfony/var-dumper": "^4.4.20 || ^5.0 || ^6.0",
```
roxblnfk, line 63:
```suggestion
        "symfony/translation": "^6.0",
        "symfony/var-dumper": "^6.0",
```

seregazhuk:
Do you really want to drop support for old symfony/laravel apps? 🤔

roxblnfk:
> Do you really want to drop support for old symfony/laravel apps? 🤔

Package dev-dependencies don't matter when you use the package in a project.
It doesn't affect anything.

PR #266, opened 2022-11-28
Fix aggregation for garbage cycles in Activity worker

roxblnfk:
## What was changed

Fix some cases when garbage cycles are got memorized.

## Why?

The cycles colud be collected with the `gc_collect_cycles` automatically but i think it will be better when we have no these ones.
It might affect some memory metrics for the better.

## Checklist

- [x] tests
- [x] fix

1. How was this tested:

I use Weak references to check the objects was destroyed after related variables unseting.


rustatian (approved):
LGTM

PR #267, opened 2022-12-11
Feature: pass error into activity finalizer

seregazhuk:
## What was changed
If an error happens when executing an activity it will be convenient to know it within a finalizer.

## Why?
For example, we collect logs and metrics and want to flush or send them only if the activity failed. Currently, within a finalizer there is no way to detect whether it was a success execution or failure.

## Checklist
1. How was this tested:
With unit tests.

3. Any docs updates needed?
Need to make a small update here with `registerActivityFinalizer()` example and provide a snippet with a closure that accepts an exception:

```md
$worker->registerActivityFinalizer(
    function (Throwable $error = null) {
        if ($error !== null) {
            $this->logger->flush();
        }
        
        $this->kernel->showtdown();
    }
);
```

 https://docs.temporal.io/application-development/foundations?lang=php#register-types

### DIFF
--- a/src/Internal/Transport/Router/InvokeActivity.php
+++ b/src/Internal/Transport/Router/InvokeActivity.php
@@ -83,7 +83,7 @@ class InvokeActivity extends Route
         } finally {
             $finalizer = $this->services->activities->getFinalizer();
             if ($finalizer !== null) {
-                call_user_func($finalizer);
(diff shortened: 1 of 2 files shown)


wolfy-j (approved):

roxblnfk (approved):

roxblnfk:
Thanks!

PR #268, opened 2022-12-16
Fix memory leak when a workflow with an active internal timer is destroyed

### LINKED ISSUE #177: [Bug] Got the response to undefined request due to the memory leak in the WF worker
### Describe the bug

Sometimes when a child worker process throws an exception, the parent worker process throws the following panic error:
```
PanicError: flush queue: SoftJobError:
	codec_execute:
	sync_worker_exec:
	sync_worker_exec_payload: LogicException: Got the response to undefined request 10389 in /srv/vendor/temporal/sdk/src/Internal/Transport/Client.php:60
```
and after it:
```
PanicError: unknown command CommandType: ChildWorkflow, ID: edfb1479-3d88-407e-a428-7e304e0d7bdf, possible causes are nondeterministic workflow definition code or incompatible change in the workflow definition 
```
Screenshot 2022-05-26 at 20 57 30

### Environment/Versions

- Temporal Version: 1.16.2 and 1.2.0 SDK
- We use Kubernetes

### Additional context

We tried to scale the pods so that can be split in different zones for fault tolerance. Maybe that causing these problems.

roxblnfk:
## What was changed

- Fixed memory leaks in the case when an uncompleted workflow with an active internal timer (see awaitWithTimeout) is destroyed.
- Call `gc_collect_cycles()` after each Workflow destroy.
- Remove the `WorkflowContext::$timers` collection. Ut is unnecessary since Temporal PHP SDK 2.0 .
- Remove the `WorkflowContext::$asyncAwaits` collection because it is unnecessary.
- Remove unused $loop parameter from the `Temporal\Internal\Transport\Client` constructor.
- Other small improvements.

## Why?

In the case when the Workflow Worker processes many workflows at the same time
and their number exceeds the value of the `cache_size` parameter, the Workflow Worker will destroy some waiting workflows.
But the Workflow Worker incorrectly destroys workflows with active internal timers: timer promises
will remain in memory. This leads to memory leak.

## Checklist

1. Closes #177
2. Autotests added

### DIFF
--- a/src/Internal/Workflow/WorkflowContext.php
+++ b/src/Internal/Workflow/WorkflowContext.php
@@ -35,7 +35,6 @@ use Temporal\Internal\Transport\Request\NewTimer;
 use Temporal\Internal\Transport\Request\Panic;
 use Temporal\Internal\Transport\Request\SideEffect;
 use Temporal\Promise;
-use Temporal\Worker\Transport\Command\CommandInterface;
 use Temporal\Worker\Transport\Command\RequestInterface;
 use Temporal\Workflow\ActivityStubInterface;
 use Temporal\Workflow\ChildWorkflowOptions;
@@ -59,12 +58,11 @@ class WorkflowContext implements WorkflowContextInterface
     protected WorkflowInstanceInterface $workflowInstance;
     protected ?ValuesInterface $lastCompletionResult = null;
 
-    protected array $awaits = [];
-    protected array $asyncAwaits = [];
     /**
-     * @var <CompletableResultInterface, CommandInterface>
+     * Contains conditional groups
+     * @var array<non-empty-string, array{callable, Deferred}>
      */
-    protected \SplObjectStorage $timers;
+    protected array $awaits = [];
 
     private array $trace = [];
     private bool $continueAsNew = false;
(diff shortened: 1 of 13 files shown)


src/Internal/Workflow/WorkflowContext.php
```diff
@@ -481,8 +501,7 @@ public function awaitWithTimeout($interval, ...$conditions): PromiseInterface
     public function resolveConditions(): void
     {
         foreach ($this->awaits as $awaitsGroupId => $awaitsGroup) {
-            foreach ($awaitsGroup as $i => $cond) {
-                [$condition, $deferred] = $cond;
+            foreach ($awaitsGroup as $i => [$condition, $deferred]) {
                 if ($condition()) {
                     $deferred->resolve();
```
wolfy-j, line 506: do we actually still need it? Maybe we can try to delete whote awaitGroup without rejecting each of internal promises (they are our promises anyway).

wolfy-j (approved):

src/Internal/Workflow/WorkflowContext.php
roxblnfk: This shuld be tested

Issue #225, opened 2022-08-03
[FEATURE REQUEST]: Add support for interceptors


bergundy:
Can use any of the other SDKs as reference.

The main uses for interceptors are context propagation, authorization and instrumentation.

PR #273, opened 2023-01-04
Header support

roxblnfk:
## What was changed

- Updated proto files to communicate with RR (added `header` field)
- Added `getHeader()` methods to Workflow and Activity facades
- Added `$header` parameter to Workflow/Activity stub factories
- Added passing for header values through Workflow/Activity executions
- Some fixes in psalm and phpunit configs

## Why?

Header supporting is required for interceptors implementation (#225)

## Additional

RoadRunner 2.12.2 is required.
Note: target branch is not the `master`


src/Workflow/WorkflowContextInterface.php
```diff
@@ -174,20 +186,31 @@ public function executeChildWorkflow(
...
+     * @param HeaderInterface|array<non-empty-string, scalar>|null $header Optional header values.
+     *        The default {@see null} value means that header values will be inherited from the current context.
+     *
      * @return T
      */
-    public function newChildWorkflowStub(string $class, ChildWorkflowOptions $options = null): object;
+    public function newChildWorkflowStub(
+        string $class,
+        ChildWorkflowOptions $options = null,
+        HeaderInterface|array|null $header = null,
```
wolfy-j, line 197: do we actually need to expose it here?

wolfy-j: can you double check with Java SDK

wolfy-j (approved):

wolfy-j:
This seems good. We have to make sure that we can properly update the SDK in relation to parent RR version.

src/Workflow/WorkflowContextInterface.php
roxblnfk: > do we actually need to expose it here?

Why not?

wolfy-j (approved):

bergundy:
We only expose headers to users via interceptors in our SDKs.
PHP shouldn't be special here.

PR #275, opened 2023-01-08
feat: unmarshalling of nested DTOs

### LINKED ISSUE #274: [Feature Request] Unmarshalling of nested DTOs
Hi!

Now it isn't possible to pass DTO containing other DTOs as properties as Workflow param.

**Example:**
```php
final class ParentDTO
{
    public function __construct(
        public readonly ChildDTO $child,
    ) {
    }
}

final class ChildDTO
{
    public function __construct(
        public readonly string $foo,
    ) {
    }
}
```

Now it causes an error:
```php
TypeError:
Cannot assign stdClass to property Temporal\Tests\Unit\DTO\DeepDTO\ParentDTO::$child of type Temporal\Tests\Unit\DTO\DeepDTO\ChildDTO
```

Slack thread [there](https://temporalio.slack.com/archives/C01LK9FAMM0/p1673178873047529).

zlodes:
## What was changed

1. In case of ObjectType: remember FQCN as `meta->of` if not set.
2. Fixed double unmarshalling in [ObjectType](https://github.com/temporalio/sdk-php/blob/12bfeda200c32fd38e0485a229f4518064d6678e/src/Internal/Marshaller/Type/ObjectType.php#L46)

## Why?
To add an ability of unmarshalling nested DTOs.

1. Closes #274 

3. How was this tested: 
Added DeepDTOTestCase

4. Any docs updates needed?
No

### DIFF
--- a/src/Internal/Marshaller/Mapper/AttributeMapper.php
+++ b/src/Internal/Marshaller/Mapper/AttributeMapper.php
@@ -153,7 +153,7 @@ class AttributeMapper implements MapperInterface
             return null;
         }
 
-        // For object-typed properties: remember property type FQCN to make object nesting possible
+        // For object-typed properties: remember property type FQN to make object nesting possible
         if ($type === ObjectType::class && $meta->of === null) {
             $meta->of = $property->getType()?->getName();
         }
--- a/src/Internal/Marshaller/Mapper/AttributeMapper.php
+++ b/src/Internal/Marshaller/Mapper/AttributeMapper.php
@@ -12,9 +12,9 @@ declare(strict_types=1);
 namespace Temporal\Internal\Marshaller\Mapper;
 
 use Spiral\Attributes\ReaderInterface;
-use Temporal\Internal\Marshaller\MarshallerInterface;
 use Temporal\Internal\Marshaller\Meta\Marshal;
 use Temporal\Internal\Marshaller\Meta\Scope;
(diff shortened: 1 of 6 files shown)


zlodes:
Sollution looks not enough clean: `AttributeMapper` shouldn't know about any specific type (`ObjectType`), but I haven't found better solution. 
Feel free to suggest. 😉

roxblnfk:
> Solution looks not enough clean: AttributeMapper shouldn't know about any specific type (ObjectType), but I haven't found better solution.

Eah. It is the main thing why I want to rewrite the existing Marshaller: type detectors have no access to type meta :|

roxblnfk (approved):

roxblnfk:
I'll continue implementation in marshaller refactoring scope

zlodes:
@roxblnfk can I help? 
It really important to have an ability to marshalling of nested typed objects in our project 😞

roxblnfk:
There is no any roadmap.
You can try to use the SDK on the `feature/marshaller` branch right now.
Your case is covered.
Also I can mention you when PR will be created.

```bash
composer require temporal/sdk "dev-feature/marshaller"
```

roxblnfk:
There are new features:

- Enums without the `#Marshal` attribute
- Readonly (thank you)
- Object types based on property type (without the `#Marshal` attribute)
- Array type works with `iterable` type

PR #278, opened 2023-01-19
Marshaller upgrade

### LINKED ISSUE #274: [Feature Request] Unmarshalling of nested DTOs

roxblnfk:
## What was changed

Using new `RuleFactoryInterface` all Marshaller types can generate Marshaller rules based on a parameter reflection.
By the reason it is unnecessary to use the `#[Marshal]` attribute in most cases. You should use the attribute
when you are using Union Types or Intersection Types.

### New features

- The `#[Marshal]` attribute now has the `$nullable` parameter. You don't need to use the `NullableType::class` value
  as the `$type` argument  inside the `#[Marshal]` attribute.  
  **Note**: `#[MarshallArray]` has different default value for the `$nullable` parameter.  
  That was made to prevent incompatible behavior (see #245).
- The `#[Marshal]` attribute can be nested. You can put another `#[Marshal]` attribute as the `$to` argument for
  `NullableType` and `ArrayType`.
- `ArrayType` can now process `iterable` type automatically.
- `ObjectType`:
  - now supports DTOs with `readonly` fields;
  - can process nested objects. Nested objects can have `#[Marshal]` attributes.
- `DateTimeType` now supports mutable and immutable DateTime types:
  - in a `DateTimeImmutable` property a `DateTimeImmutable` object will be unmarshalled;
  - in a `DateTime` property a `DateTime` object will be unmarshalled;
  - in a `Carbon\CarbonImmutable` property a `Carbon\CarbonImmutable` object will be unmarshalled;
  - in a `DateTimeInterface` or `Carbon\Carbon` property a `Carbon\Carbon` object will be unmarshalled;
  - to change default behavior you can pass to the `#[Marshal]` attribute the needed class name as the `$of` argument.

### Changes and deprecations:

- The method `ObjectType::instance()` has been marked as deprecated.
- Marshaller will throw a detailed exception about invalid argument value during unmarshalling.
- > **Warning**: marshalling rules that was got from the `#[Marshal]` attribute now are stricter than before.  
  > All attributed properties with nullable type will be not nullable by default (excluding `#[ArrayType]`).  
  > In the case when the `null` value is passed, an exception will be thrown.  
  > In this case you just need to remove the `#[Marshal]` attribute or set the `$nullable` parameter to `true`.

## Checklist

1. Close #274
2. How was this tested: Unit tests
3. Documentation changes: there are no documentation about Mershaller in the Temporal docs.


roxblnfk:
@zlodes hi. Do you want to test it and send a feedback?

wolfy-j (approved):

zlodes:
@roxblnfk Hi! Yeah. I'm going to test it in a couple of days. 
Thanks!

roxblnfk:
> @roxblnfk Hi! Yeah. I'm going to test it in a couple of days. Thanks!

That's good. Plese give a feedback here after that. Thank you.

To test you should use `dev-master`.

`composer require temporal/sdk:dev-master`

zlodes:
@roxblnfk hi! Now Marshaller works fine with nested DTO. Thanks! 🚀 

I have a question connected to Marshaller.

Now I'm getting an error about marshalling of enums. 
I know that it won't work by default due to support of oldest PHP versions. 

Also we use [Decimal extension](https://github.com/php-decimal/ext-decimal). 
I added a `DecimalType` class extends abstract `Type` and use it with `#[Marshal]` attribute. 

Is there a way to use Marshaller w/o `#[Marshal]` attribute for these cases?

roxblnfk:
@zlodes thank you for feedback!

> Now I'm getting an error about marshalling of enums.
> I know that it won't work by default due to support of oldest PHP versions.

PHP 8.1 enums will work without attribute if you find a way to create and inject [the Marshaller](https://github.com/temporalio/sdk-php/blob/fe0c2e0baca6c77ed636b20e07311014cdf4b418/src/Internal/Marshaller/Marshaller.php#L42) with the type [EnumType](https://github.com/temporalio/sdk-php/blob/fe0c2e0baca6c77ed636b20e07311014cdf4b418/src/Internal/Marshaller/Type/EnumType.php#L17) in the second argument.

I think I should make it easier and add the EnumType by default for PHP >8.1 🤔.
@wolfy-j WDYT?

zlodes:
@roxblnfk Yeap, I think it would be better to add `EnumType` to `TypeFactory::getDefaultMatchers`

PR #282, opened 2023-02-01
Add EnumType by default for PHP >= 8.1

### LINKED ISSUE #281: Add `EnumType` by default for PHP 8.1+
I think we should add the `EnumType` by default for PHP >=8.1

_Originally posted by @roxblnfk in https://github.com/temporalio/sdk-php/issues/278#issuecomment-1411790904_

zlodes:
Closes #281

### DIFF
--- a/src/Internal/Marshaller/TypeFactory.php
+++ b/src/Internal/Marshaller/TypeFactory.php
@@ -15,6 +15,7 @@ use Temporal\Internal\Marshaller\Type\ArrayType;
 use Temporal\Internal\Marshaller\Type\DateIntervalType;
 use Temporal\Internal\Marshaller\Type\DateTimeType;
 use Temporal\Internal\Marshaller\Type\DetectableTypeInterface;
+use Temporal\Internal\Marshaller\Type\EnumType;
 use Temporal\Internal\Marshaller\Type\ObjectType;
 use Temporal\Internal\Marshaller\Type\RuleFactoryInterface as TypeRuleFactoryInterface;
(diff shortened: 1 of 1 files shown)


src/Internal/Marshaller/TypeFactory.php
```diff
@@ -138,5 +139,9 @@ private function getDefaultMatchers(): iterable
         yield DateIntervalType::class;
         yield ArrayType::class;
         yield ObjectType::class;
+
+        if (PHP_VERSION_ID >= 80104) {
+            yield EnumType::class;
+        }
```
roxblnfk, line 145: It preffer to place it before the `ObjectType`.
Please move the block to the top

roxblnfk: It need to cover by a test also

roxblnfk: I'll do it.
I've detected some bugs in EnumType and i'll fix it.

roxblnfk:
Fixed there https://github.com/temporalio/sdk-php/pull/284

@zlodes hi! What about testing for enums without attributes in your project? :)

zlodes:
@roxblnfk hi! Thanks!

Now it works great w/o `#Marshal` attributes! 🚀

Marshaller class seems to be a service (singleton). 
I'll think about an ability to extend available types with custom ones (like `DecimalType` from my example`).

roxblnfk:
Thanks.
Let me know if there is a legal way to add custom type.
If it is impossible then we will improve that.

PR #298, opened 2023-04-17
Expand list of retryable gRPC errors

### LINKED ISSUE #296: Include protocol errors into retryable statements
### Describe the bug

protocol errors are not being retried (only service level errors)  https://github.com/temporalio/sdk-php/blob/6bc84c0d931f6187d72566a6ce134b5476799e87/src/Client/GRPC/BaseClient.php#L137-L139 , please retry protocol errors too. 

thanks, 

### Environment/Versions

- OS and processor: [e.g. M1 Mac, x86 Windows, Linux]
- Temporal Version: [e.g. 1.14.0?] and/or SDK version
- Are you using Docker or Kubernetes or building Temporal from source?

roxblnfk:
## What was changed

UNAVAILABLE and UNKNOWN gRPC errors now are retryable.

> **Note**:
> In the case when a Temporal server is unavailable and by the reason a Client call can't be completed, without a limited retry gRPC call policy it might retry infinitely.

## Why?

In some cases gRPC connection might be broken or not initialized. In this case we should follow our retry policy.

## Checklist

1. Closes #296 

2. How was this tested:
Manually using [proxy](https://github.com/Shopify/toxiproxy)

### DIFF
--- a/src/Client/GRPC/BaseClient.php
+++ b/src/Client/GRPC/BaseClient.php
@@ -139,7 +139,6 @@ abstract class BaseClient implements ServiceClientInterface
                 [$result, $status] = $call->wait();
 
                 if ($status->code !== 0) {
-                    echo $status->code . "\n";
                     throw new ServiceClientException($status);
                 }
 
--- a/src/Client/GRPC/BaseClient.php
+++ b/src/Client/GRPC/BaseClient.php
@@ -21,7 +21,6 @@ abstract class BaseClient implements ServiceClientInterface
 {
     const RETRYABLE_ERRORS = [
         StatusCode::RESOURCE_EXHAUSTED,
-        StatusCode::INTERNAL,
         StatusCode::UNAVAILABLE,
         StatusCode::UNKNOWN,
     ];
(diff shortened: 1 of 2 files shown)


wolfy-j (approved):

wolfy-j:
Can you please include status code into error message?

src/Client/GRPC/BaseClient.php
```diff
@@ -19,6 +19,11 @@
 
 abstract class BaseClient implements ServiceClientInterface
 {
+    const RETRYABLE_ERRORS = [
+        StatusCode::RESOURCE_EXHAUSTED,
+        StatusCode::UNAVAILABLE,
+        StatusCode::UNKNOWN,
+    ];
```
Sushisource, line 26: We ought to expand this to include the ones here https://github.com/temporalio/sdk-core/blob/81ece42e3a4fda1b019d9a5d8d263916f73726a6/client/src/retry.rs#LL22C1-L22C1

Sushisource: We'll also want to double check that the retry logic itself is the same - using exponential backoff etc. The logic is described later in the file here: https://github.com/temporalio/sdk-core/blob/81ece42e3a4fda1b019d9a5d8d263916f73726a6/client/src/retry.rs#L182

DM me if you have any questions, thanks!

PR #287, opened 2023-02-05
Add PHP 8.2 in CI tests

roxblnfk:
## What was changed

Added PHP 8.2 in CI tests

PR #299, opened 2023-04-17
[CI] Add Windows tests

roxblnfk:
## What was changed

Add Windows tests in CI

## Why?

Windows tests was removed with #287 because there are [no gRPC binaries in PECL repository](https://pecl.php.net/package/grpc) for PHP 8.2

### Related issues:
- https://github.com/php/php-src/issues/10850
- https://github.com/grpc/grpc/issues/32304

### Links:
- Related thread [Windows PECL build machine died](https://externals.io/message/117494) on PHP Externals

### DIFF
--- a/.github/workflows/run-test-suite.yml
+++ b/.github/workflows/run-test-suite.yml
@@ -27,8 +27,11 @@ jobs:
       fail-fast: ${{ inputs.fail-fast }}
       matrix:
         php: [ 8.1, 8.2 ]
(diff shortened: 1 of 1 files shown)


roxblnfk:
**Note**:

We can use that code in the matrix
```
        exclude:
          - os: windows-latest
            php: 8.2
```

to disable PHP 8.2 on Windows

wolfy-j (approved):

PR #306, opened 2023-05-26
Interceptors

roxblnfk:
> **Note**:
> The PR is opened into `2.6.0` branch, not the `master`.
> It means after merge users will be able to install the SDK 2.6.0 version with the `minimum-stabiliy: dev` flag.

## What was changed

The PR includes #273, #293, #295, #301, #303, #305 and new features.

There are few interceptors that we may add later:

- New `TransportInterceptor` - intercepts all requests/responses between SDK-PHP and RoadRunner
- New `SystemInfoInterceptor` - like in the Java SDK
- `WorkflowOutboundCallsInterceptor` add methods:
- `registerDeclaredQueries()` - initial query handlers registration
  - `registerDeclaredSignals()` - initial signal handlers registration
  - `registerQuery()` - dynamic query handler registration
  - `registerSignal()` - dynamic signal handler registration
- `WorkflowInboundInterceptor`: add `CancelWorkflow()` method

> **Note**:
> `OpenTelemetryInterceptor`s are in [this separated repo](https://github.com/temporalio/sdk-php-interceptors-opentelemetry)

## Why?

Because of #225

## Checklist

1. How was this tested: autotests and manual tests
2. Any docs updates needed? - yes

> **Note**:
> Psalm issues will be fixed during testing period.


src/Client/GRPC/BaseClient.php
```diff
@@ -114,6 +131,25 @@ protected function invoke(string $method, object $arg, ContextInterface $ctx = n
     {
         $ctx = $ctx ?? Context::default();
 
+        return $this->invokePipeline !== null
+            ? ($this->invokePipeline)($method, $arg, $ctx)
+            : $this->call($method, $arg, $ctx);
+    }
+
+    /**
+     * Call grpc.
```
Sushisource, line 140:
```suggestion
     * Call a grpc method.
```

src/Interceptor/HeaderInterface.php
```diff
@@ -0,0 +1,60 @@
...
+     * Make a protobuf Header message.
+     *
+     * @return Header
+     */
+    public function toHeader(): Header;
+
+    /**
+     * @param DataConverterInterface $converter
+     */
+    public function setDataConverter(DataConverterInterface $converter);
```
Sushisource, line 59: It's odd that you can set a data converter specifically for headers. We typically allow data converters to be set on clients and/or workers, and sometimes around a specific command in a workflow, but it's odd to have it publicly exposed that a converter could be set for a specific header

src/Interceptor/WorkflowClient/QueryInput.php
```diff
@@ -0,0 +1,58 @@
...
+        #[Immutable]
+        public ?string $workflowType,
+        #[Immutable]
+        public string $queryType,
+        #[Immutable]
+        public ValuesInterface $arguments,
+        // #[Immutable]
+        // public Class<R> $resultClass,
+        // #[Immutable]
+        // public Type $resultType,
```
Sushisource, line 38: Dead comments? Here and below

Sushisource (approved):
This looks like it all makes sense to me. Thanks! I wasn't able to give the most in-depth review ever, but this is all fairly straightforward.

I really only have one question around allowing converters to be explicitly specified on a per-header basis. If that isn't actually publicly exposed, then no need to worry about it.

src/Interceptor/HeaderInterface.php
roxblnfk: In this PR I'll mark it `@internal` and I might remove it in the next PR.

The way it's done now is that headers are not decoded if they're not requested, and are sent as is if they haven't been changed. And those values that were added through `withValue()` are encoded into Payload before sending.

I'll try to do it differently: headers on input will be decoded immediately and packed again only before sending.

OK?

Sushisource: Yeah, I think that makes good sense

PR #307, opened 2023-05-29
Adding RoadRunner version checker

msmakouz:
## What was changed

Added installed RoadRunner version checking. Min PHP version 8.1.

## Why?

Interceptors requires a minimum RR version ^2023.1


roxblnfk:
FYI
Around 10% users use PHP < 8.1.

img

https://packagist.org/packages/temporal/sdk/php-stats

PR #315, opened 2023-07-04
Add ability to get history length from WorkflowInfo

### LINKED ISSUE #222: [Feature Request] Expose history length via Workflow info

roxblnfk:
## What was changed

Added `WorkflowInfo::$historyLength` property.
This field will be updated with any response from the RoadRunner server.

> **Note**:
> The feature is proposed into Interceptors branch because there will be a lot of conflicts if we merge it into master branch before the Interceptors branch is merged.

> **Note**:
> The ability will be available from [RoadRunner v2023.2](https://github.com/roadrunner-server/roadrunner/releases/tag/v2023.2.0) (https://github.com/temporalio/roadrunner-temporal/pull/382).
> With lower RR versions, this field is always 0.
> Waiting for https://github.com/temporalio/sdk-php/pull/307. Then we will have the latest RR version in the tests

## Why?

Users should be able to have a more direct way to check if it's time to call continue-as-new than guessing how many events their code produces and counting the loops.

## Checklist

- Closes #222 
How was this tested: 
- [x] Added autotest.
- [x] Manually tested with different RoadRunner versions.
- [x] Test deterministic behavior: it has to be exactly the same during a replay.


src/Exception/Failure/FailureConverter.php
```diff
@@ -206,39 +205,36 @@ private static function createFailureException(Failure $failure, DataConverterIn
...
+                    : EncodedValues::fromPayloads($info->getDetails(), $converter);
 
                 return new CanceledFailure($failure->getMessage(), $details, $previous);
 
-            case $failure->hasTerminatedFailureInfo():
+            case $failure->getTerminatedFailureInfo() !== null:
                 return new TerminatedFailure($failure->getMessage(), $previous);
 
-            case $failure->hasServerFailureInfo():
+            case $failure->getServerFailureInfo() !== null:
```
roxblnfk, line 229: There are a lot of changes like this.
That related with the fact I used an old Protoc version and it removed all `has*` methods after compilation.
I think I should revert it.

roxblnfk:
> **Note**:
> `master` has been merged into this branch

PR #316, opened 2023-07-04
Adding Marshaller UuidType

### LINKED ISSUE #317: Support UUID in marshaller

Users use UUID a lot: generate like a side effect, pass in activities, get as results.
Right now users must cast UUID objects to string manually.

It will be nice to support UUID objects in our marshaller.

We have the great UUID library (https://github.com/ramsey/uuid) in PHP world.
There is also `symfony/uid` but it is not a popular package yet.

I think we can support the first or both in the SDK. We can do it in optional way - if a UUID library wasn't installed in the project then related typecaster will be deactivated.

msmakouz:
## What was changed

Added new **Temporal\Internal\Marshaller\Type\UuidType**.

## Why?

This type allows the uuid to be used in classes that are serialized by the Marshaller.

## Checklist

- Fix #317

### DIFF
--- a/src/Internal/Marshaller/Type/UuidType.php
+++ b/src/Internal/Marshaller/Type/UuidType.php
@@ -18,13 +18,17 @@ use Temporal\Internal\Support\Inheritance;
 
 final class UuidType extends Type implements DetectableTypeInterface, RuleFactoryInterface
 {
+    private static ?bool $interfaceExists = null;
+
     public static function match(\ReflectionNamedType $type): bool
     {
-        if (!\interface_exists(UuidInterface::class)) {
-            return false;
+        if (self::$interfaceExists === null) {
+            self::$interfaceExists = \interface_exists(UuidInterface::class);
         }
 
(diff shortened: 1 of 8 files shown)


src/Internal/Marshaller/Type/UuidType.php
```diff
@@ -0,0 +1,62 @@
...
+use Ramsey\Uuid\Uuid;
+use Ramsey\Uuid\UuidInterface;
+use Temporal\Internal\Marshaller\MarshallingRule;
+use Temporal\Internal\Support\Inheritance;
+
+final class UuidType extends Type implements DetectableTypeInterface, RuleFactoryInterface
+{
+    public static function match(\ReflectionNamedType $type): bool
+    {
+        if (!\interface_exists(UuidInterface::class)) {
```
roxblnfk, line 23: It might be better to cached it in a static variable here

msmakouz: Added static variable.

PR #321, opened 2023-07-04
[Feature] Implement UUID conversion

### LINKED ISSUE #317: Support UUID in marshaller

msmakouz:
## What was changed

Added new **Temporal\Internal\Marshaller\Type\UuidType**.

## Why?

This type allows the uuid to be used in classes that are serialized by the Marshaller.

## Checklist

Fix #317


composer.json
```diff
@@ -61,7 +61,8 @@
         "phpunit/phpunit": "^9.5.21",
         "symfony/translation": "^6.0",
         "symfony/var-dumper": "^6.0",
-        "vimeo/psalm": "^4.1"
+        "vimeo/psalm": "^4.1",
+        "ramsey/uuid": "4.7"
```
roxblnfk, line 65:
```suggestion
        "ramsey/uuid": "^4.7"
```

roxblnfk (approved):

composer.json
msmakouz: Fixed, thanks

PR #324, opened 2023-07-05
[Feature] Friendly version of listWorkflowExecutions

roxblnfk:
## What was changed

Added the `listWorkflowExecutions()` method to `WorkflowClientInterface`.
User can now query for workflow executions using a list filter query syntax.
As result the Paginator object is returned which can be used in a simple way:

```php
$paginator = $client->listWorkflowExecutions(
    "WorkflowType='MyWorkflow' and StartTime  between '2022-08-22T15:04:05+00:00' and  '2023-08-22T15:04:05+00:00'",
    'default',
    10,
);

// Iterate all items (pages will be fetched automatically)
foreach ($paginator as $execution) {
    // ...
}

// Get current page items
$items = $paginator->getPageItems();

// Get next page
$nextPage = $paginator->getNextPage();
```

I can add `CountWorkflowExecutionsRequest` call there (inside the Paginator) as well if needed.

## Checklist

- PHP part of https://github.com/temporalio/features/issues/75
- How was this tested:
  - Tested manually
  - Added autotests
- Any docs updates needed?
  - I'm not sure


roxblnfk:
@Sushisource please validate these data objects:

[src/Client/DTO/WorkflowExecutionInfo.php](https://github.com/temporalio/sdk-php/pull/324/files#diff-cb29d6754ad861e9e16b0e637c7c8ac2554537a081b1571ef490272f6045c3f6)
[src/Client/DTO/ResetPointInfo.php](https://github.com/temporalio/sdk-php/pull/324/files#diff-4c89936cd695dc53c5b7091f580827ddc33ec8afad9ea710e027119795c71a17)

---

> I can add CountWorkflowExecutionsRequest call there (inside the Paginator) as well if needed.

Also, should I make Paginator countable?

wolfy-j (approved):

src/Client/DTO/WorkflowExecutionInfo.php
```diff
@@ -0,0 +1,41 @@
...
+{
+    /**
+     * @param array<ResetPointInfoDto> $autoResetPoints
+     */
+    public function __construct(
+        public WorkflowExecution $execution,
+        public WorkflowType $type,
+        public ?DateTimeInterface $startTime,
+        public ?DateTimeInterface $closeTime,
+        public int $status,
```
Sushisource, line 29: The status here should be some kind of well-defined enum based on https://github.com/temporalio/api/blob/5f21c72de0ec9e6037e84c95f097641b71d7c39f/temporal/api/enums/v1/workflow.proto#L74

src/Client/DTO/WorkflowExecutionInfo.php
```diff
@@ -0,0 +1,41 @@
...
+        public int $historyLength,
+        public ?string $parentNamespaceId,
+        public ?WorkflowExecution $parentExecution,
+        public ?DateTimeInterface $executionTime,
+        public EncodedCollection $memo,
+        public EncodedCollection $searchAttributes,
+        public array $autoResetPoints,
+        public string $taskQueue,
+        public int $stateTransitionCount,
+    ) {
```
Sushisource, line 39: This is also missing the recently added version stamp: https://github.com/temporalio/api/blob/5f21c72de0ec9e6037e84c95f097641b71d7c39f/temporal/api/workflow/v1/message.proto#L61

Sushisource (commented):
> Also, should I make Paginator countable?

By this do you mean expose some function on it that will go through all the pages and return a count? Is it [this](https://www.php.net/manual/en/class.countable.php)? (Forgive my lack of php knowledge).

If it means implementing that interface, I think that's fine - of course it means it's going to consume all the pages and thus either need to store everything in memory, or become unusable after `count`ing. I don't think the former is a good idea. The latter is fine, as long as there's a clear way to express the exhaustion of the iterator if a user tries to use it after `count`ing.

roxblnfk:
> If it means implementing that interface, I think that's fine - of course it means it's going to consume all the pages and thus either need to store everything in memory, or become unusable after `count`ing. I don't think the former is a good idea. The latter is fine, as long as there's a clear way to express the exhaustion of the iterator if a user tries to use it after `count`ing.

I planned to get the items count using the `CountWorkflowExecutionsRequest` RPC request under the hood.

Sushisource:
@roxblnfk Ah, in that case, yeah I think that's totally fine.

src/Client/DTO/WorkflowExecutionInfo.php
roxblnfk: PHP Enums are available since PHP 8.1.
PHP 8.1 is required by interceptors.
We can bump PHP version right now to ^8.1, or use constants and it will look like deprecation after the interceptors release

Sushisource: Ah, I see. I don't have a sense of how disruptive the PHP version bump would be. If that's unlikely to be a big problem for anyone I'd say let's go for that, otherwise I'm fine leaving it a scalar (with predefined constants) for now.

roxblnfk: Any way we have to up to 8.1 with the interceptors release

Sushisource (approved):

PR #331, opened 2023-07-12
Fix exceptions propagation from Signal methods

roxblnfk:
## What was changed

Exceptions from Signal methods now will be propagated and affect Workflow.
Deserialization errors will be ignored. Called Signal method in this case will be skipped.

## Checklist

How was this tested: manually; also added autotests.

> **Note**:
> The fix will be cherry-picked into `master` and `2.7` branches after `2.5.3` release

### DIFF
--- a/src/Internal/Workflow/Process/Scope.php
+++ b/src/Internal/Workflow/Process/Scope.php
@@ -180,6 +180,17 @@ class Scope implements CancellationScopeInterface, PromisorInterface
     }
 
     /**
+     * @param callable $handler
+     */
+    public function startSignal(callable $handler): void
+    {
+        // Create a coroutine generator
+        $this->coroutine = $this->callSignalHandler($handler);
+        $this->context->resolveConditions();
+        $this->next();
+    }
+
+    /**
      * @param \Generator $generator
      * @return self
      */
@@ -326,8 +337,32 @@ class Scope implements CancellationScopeInterface, PromisorInterface
     {
         try {
             $this->makeCurrent();
+            $result = $handler($values);
+
+            if ($result instanceof \Generator) {
+                yield from $result;
+
+                return $result->getReturn();
+            }
+
+            return $result;
+        } catch (\Throwable $e) {
+            $this->onException($e);
+        }
+    }
+
+    /**
+     * Call a Signal method. In this case deserialization errors are skipped.
+     *
+     * @param callable $handler
+     * @return \Generator
+     */
+    protected function callSignalHandler(callable $handler): \Generator
+    {
+        try {
+            $this->makeCurrent();
             try {
-                $result = $handler($values);
+                $result = $handler();
             } catch (InvalidArgumentException) {
                 // Skip deserialization errors
                 return null;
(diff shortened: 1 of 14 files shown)


wolfy-j (approved):

tests/Functional/Client/FailureTestCase.php
```diff
@@ -80,4 +83,66 @@ public function testChildWorkflowFailurePropagation()
...
+        $run = $client->startWithSignal($wf, 'failActivity', ['foo']);
+
+        try {
+            sleep(5);
+            $wf->failActivity('foo');
+            $this->fail('Signal must fail');
+        } catch (AssertionFailedError $e) {
+            throw $e;
+        } catch (\Throwable $e) {
+            $this->assertInstanceOf(WorkflowNotFoundException::class, $e);
```
Sushisource, line 141: Hmm... is there an `ActivityNotFoundException`? Seemingly this should throw that?

Sushisource (approved):
Approving since I'm pretty sure this makes sense, but, a couple things to double-check.

tests/Functional/Client/FailureTestCase.php
wolfy-j: Not really, this error is when we are trying to send a signal to an already dead workflow.

tests/Functional/Client/AwaitTestCase.php
```diff
@@ -133,7 +133,16 @@ public function testFailSignalSerialization()
...
 
         // breaks the invocation
-        $wait->addValue(['hello'], 123);
+        //
+        // Throws:
+        // Temporal\Exception\Failure\ApplicationFailure
+        // Previous:
+        // Temporal\Exception\InvalidArgumentException:
+        // The passed value of type "array" can not be converted to required type "string" in
+        // src\Internal\Declaration\Dispatcher\AutowiredPayloads.php:34
```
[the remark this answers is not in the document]
wolfy-j: OK, we can make it less strict. Thx for responding.

tests/Functional/Client/FailureTestCase.php
Sushisource: Oooh, I see, I missed that it's already start-with-signal failing the activity earlier. Got it.

tests/Functional/Client/AwaitTestCase.php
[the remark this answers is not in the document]
roxblnfk: Deserialization errors will be ignored. Called Signal method in this case will be skipped.

src/Internal/Workflow/Process/Scope.php
```diff
@@ -322,16 +324,25 @@ function () use ($cancelID): void {
...
     protected function call(callable $handler, ValuesInterface $values): \Generator
     {
-        $this->makeCurrent();
-        $result = $handler($values);
+        try {
+            $this->makeCurrent();
+            try {
+                $result = $handler($values);
+            } catch (InvalidArgumentException) {
+                // Skip deserialization errors
```
Sushisource, line 332: Just want to confirm there aren't any other exception types that could result from bad deserialization that need to be included here.

Sushisource (approved):

src/Internal/Workflow/Process/Scope.php
roxblnfk: As you can see it throws SDK's `InvalidArgumentException` only
https://github.com/temporalio/sdk-php/blob/cf2f921a6ca923f7e6ccffc3d12314c445af236a/src/Internal/Declaration/Dispatcher/AutowiredPayloads.php#L30-L36

wolfy-j (approved):

PR #333, opened 2023-07-17
Add API to count workflows

### LINKED ISSUE #291: Add API to count workflows
See parent issue for more information

roxblnfk:
## What was changed

Added `WorkflowClient::countWorkflowExecutions()` method

## Checklist

1. Closes #291
3. Any docs updates needed? I think the block [Observability / Visibility](https://docs.temporal.io/dev-guide/php/observability#visibility) may be updated

### DIFF
--- a/src/Client/WorkflowClient.php
+++ b/src/Client/WorkflowClient.php
@@ -277,7 +277,7 @@ class WorkflowClient implements WorkflowClientInterface
                 $request->setNextPageToken($nextPageToken);
             } while ($nextPageToken !== '');
         };
-        $counter = fn(): int => (int)$this->countWorkflowExecutions($query, $namespace)->getCount();
+        $counter = fn(): int => $this->countWorkflowExecutions($query, $namespace)->count;
 
         return Paginator::createFromGenerator($loader($request), $counter);
     }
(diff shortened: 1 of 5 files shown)


src/Client/WorkflowClient.php
```diff
@@ -276,17 +276,28 @@ public function listWorkflowExecutions(
...
         return Paginator::createFromGenerator($loader($request), $counter);
     }
 
+    /**
+     * @inheritDoc
+     */
+    public function countWorkflowExecutions(
+        string $query,
+        string $namespace = 'default',
+    ): int {
```
cretz, line 290: This should not return an int as we have new result info coming at https://github.com/temporalio/api/pull/303 that can returned grouped counts

roxblnfk: Got it. Will be a DTO

src/Client/WorkflowClient.php
```diff
@@ -276,17 +277,24 @@ public function listWorkflowExecutions(
...
         return Paginator::createFromGenerator($loader($request), $counter);
     }
 
+    /**
+     * @inheritDoc
+     */
+    public function countWorkflowExecutions(
+        string $query,
+        string $namespace = 'default',
+    ): CountWorkflowExecutionsResponse {
```
cretz, line 291: IMO you should return a high-level wrapper around the response instead of the raw gRPC response. In Go we regret using raw proto responses and in other SDKs we do not.

wolfy-j: Ugh, i specifically decided to go with GRPC one since it contains a single value. @roxblnfk sorry about that.

Sushisource (approved):

PR #336, opened 2023-08-02
Expose Replayer API

### LINKED ISSUE #227: Expose Replayer API
See other SDKs for reference.
Go SDK already provides this functionality, need to wrap the API and expose to PHP.
The purpose of the replayer is:
- replay workflow histories against new workflow code to check for non-determinism
- replay workflow history from a production workflow and debug locally

roxblnfk:
> **Note**: 
> Should be tested with any nightly or stable build of RoadRunner 2023.3
> All tests are passed successfully but locally and on nightly build

> **Note**:
> The PR contains fixes from the `2.5.x`.

## What was changed

Deleted generated protobuf files and related git submodule because they are provided by `roadrunner-php/roadrunner-api-dto`.

Added `WorkflowReplayer` class.
The class exposes the following methods:
- `replayFromServer()` - Replays workflow from a history that will be loaded from Temporal server.
- `downloadHistory()` - Downloads workflow history from Temporal server and saves it to a file.
- `replayFromJSON()` - Replays workflow from a json serialized history file.
- `replayHistory()` - Replays History from a protobuf object (`temporal.api.history.v1.History`)

It uses RoadRunner RPC to communicate with Temporal server.

Added the method `\Temporal\Client\WorkflowClientInterface::getWorkflowHistory()` that can load history in PHP memory.

## Why?

Replayer API is very useful tool for a workflow determinism testing.

## Checklist

- Closes #227
- RoadRunner Temporal plugin RFC https://github.com/roadrunner-server/roadrunner/issues/1640
- [x] Autotests
- [x] Documentation
- [x] Merge and release https://github.com/roadrunner-php/roadrunner-api-dto/pull/2
- [x] Update and release `roadrunner-php/roadrunner-api-dto` with new DTO
- [x] Samples https://github.com/temporalio/samples-php/pull/38
- [ ] Release RR 2023.3

--- 

# Documentaion

## How to Replay a Workflow Execution

Replay recreates the exact state of a Workflow Execution. You can replay a Workflow from the beginning of its Event History.

Replay succeeds only if the [Workflow Definition](https://docs.temporal.io/workflows#workflow-definition) is compatible with the provided history from a deterministic point of view.

When you test changes to your Workflow Definitions, we recommend doing the following as part of your CI checks:

1. Determine which Workflow Types or Task Queues (or both) will be targeted by the Worker code under test.
2. Download the Event Histories of a representative set of recent open and closed Workflows from each Task Queue, either programmatically using the SDK client or via `tctl`.
3. Run the Event Histories through replay.
4. Fail CI if any error is encountered during replay.

The following are examples of fetching and replaying Event Histories:

To replay Workflow Executions, use the `\Temporal\Testing\Replay\WorkflowReplayer` class.

In the following example, Event Histories are downloaded from the server, and then replayed. Note that this requires Advanced Visibility to be enabled.

```php
/**
 * We assume you already have a WorkflowClient and WorkflowReplayer in scope.
 * @var \Temporal\Client\WorkflowClientInterface $workflowClient
 * @var \Temporal\Testing\Replay\WorkflowReplayer $replayer
 */

// Find all workflow executions of type "MyWorkflow" and task queue "MyTaskQueue".
$executions = $workflowClient->listWorkflowExecutions(
    "WorkflowType='MyWorkflow' AND TaskQueue='MyTaskQueue'"
);

// Replay each workflow execution.
foreach ($executions as $executionInfo) {
    try {
        $replayer->replayFromServer(
            workflowType: $executionInfo->type->name,
            execution: $executionInfo->execution,
        );
    } catch (\Temporal\Testing\Replay\Exception\ReplayerException $e) {
        // Handle the replay error.
    }
}
```

In the next example, a single history is loaded from a JSON file on disk:

```php
$replayer->replayFromJSON(
    workflowType: 'MyWorkflow',
    filePath: 'history.json',
    lastEventId: 42,  // optional
);
```

In both examples, if Event History is non-deterministic, an error is thrown.

You can download a workflow history using PHP and replay it from a memorized History object:

```php
$history = $this->workflowClient->getWorkflowHistory(
    execution: $run->getExecution(),
)->getHistory();

(new WorkflowReplayer())->replayHistory($history);
```


testing/src/Replay/Exception/NonDeterministicWorkflowException.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Temporal\Testing\Replay\Exception;
+
+final class NonDeterministicWorkflowException extends ReplayerException
+{
+}
```
Sushisource, line 9: It's worth having a docstring here that links to  https://docs.temporal.io/workflows/#deterministic-constraints

testing/src/Replay/WorkflowReplayer.php
```diff
@@ -4,20 +4,138 @@
...
+    private RPC $rpc;
+
+    public function __construct()
+    {
+        $this->rpc = new RPC(Relay::create(Environment::fromGlobals()->getRPCAddress()), new ProtobufCodec());
+    }
+
     /**
-     * Replays workflow from a resource that contains a json serialized history.
+     * Replays workflow from a history that will be loaded from Temporal server.
```
Sushisource, line 36:
```suggestion
     * Replays a workflow from history that will be fetched from Temporal server.
```

testing/src/Replay/WorkflowReplayer.php
```diff
@@ -4,20 +4,138 @@
...
+
+    /**
+     * Downloads workflow history from Temporal server and saves it to a file.
+     *
+     * @param non-empty-string $workflowType
+     * @param non-empty-string $savePath
+     *
+     * @throws ReplayerException
+     */
+    public function downloadHistory(
```
Sushisource, line 56: This should download the history and return it as a proto, I think. Then we should also provide a function for saving protos to a JSON file. This way the user can do everything in memory without touching disk, and enables inspecting history before replaying if they would like to (which they can't with `replayFromServer`). 

Semi-related: Where are the implementations of the other side of these `sendRequest` calls? I can't find the corresponding definitions.

wolfy-j: This is RR server side. https://github.com/temporalio/roadrunner-temporal/pull/394/files#diff-05a67dd647cb7c91ba1c5f7d5382094b4b839d86ca8d4e90e288f0566329c76cR188

roxblnfk: > Where are the implementations of the other side of these sendRequest calls? I can't find the corresponding definitions.

`$this->rpc->call($commad, $request);` does RoadRunner RPC call.
RFC about Replayer API there: https://github.com/roadrunner-server/roadrunner/issues/1640

Sushisource: Ah, cool, I figured that was the case I just didn't realize I should go look in that other repo. Makes sense.

Sushisource: Yes, that makes good sense.

src/Client/WorkflowExecutionHistory.php
```diff
@@ -71,4 +71,22 @@ public function getIterator(): Traversable
...
+            throw new \RuntimeException(\sprintf('Directory "%s" does not exist.', \dirname($file)));
+        }
+
+        $events = \iterator_to_array($this->getEvents(), false);
+        $history = new History();
+        $history->setEvents($events);
+
+        $data = $history->serializeToJsonString();
+        \file_put_contents($file, $data);
+    }
```
roxblnfk, line 91: @Sushisource hi.
I'm trying to store Events History using protobuf to json serialization.

And I'm getting invalid JSON as result:

[history-files.zip](https://github.com/temporalio/sdk-php/files/12293461/history-files.zip)

All enums have different values:
{9BE90F8F-C69D-4797-AA3E-B1C258BEB670}

If we really need to have the ability to store history from PHP collection then I need to convert all these values in a right way.

wolfy-j: Well, at least properties named correctly.

Sushisource: Yeah, so you'll need to use the Go SDK to deserialize such histories, because it uses a weird forked version of gogoproto that does this. It's quite obnoxious. I was thinking you were already going through Go for both the serialize and deserialize operations.

We could do something annoying like check the first event (always workflow execution started) and depending on how `eventType` looks, either deserialize via Go, or do it directly.

This would result in supporting downloaded histories from the Web UI too, which also get serialized the Go way, so that is probably desirable. Annoying, but should be fairly easy.

wolfy-j: We have already implemented the whole history download in both formats and replay on Golang's end. We have tested the ability to replay history from UI with the code that was exposed with these 3 methods.

Sounds like reading history on PHP is not a problem, we can probably also save it to protobuf file and replay, but I'm not sure why we need to convert it anyhow when all the de-serialization logic is on Golang end.

wolfy-j: We can also try to de-serialize whole history from pb to JSON using new RPC command instead of going entry by entry.

roxblnfk:
I updated PR description.
Now we can replay any workflow from a History protobuf object.

wolfy-j (approved):

testing/src/Replay/WorkflowReplayer.php
```diff
@@ -33,7 +39,36 @@ public function __construct()
...
+     * Replays a workflow from {@see History}
+     *
+     * @param non-empty-string|null $workflowType Workflow type name. If not specified, the workflow type will be
+     *        extracted from the first event in the history.
+     *
+     * @throws ReplayerException
+     */
+    public function replayHistory(
+        History $history,
+        ?string $workflowType = null,
```
Sushisource, line 51: In most of the other SDKs we allow specifying the workflow id (since, oddly, that's not in history) rather than the type, which there isn't really much reason to specify since it ought to be extracted from the first event as you've done here.

If the user wants to try replaying a different wf type against that history, they simply can just register the workflow function with the same type name as is present in the history

wolfy-j: The workflow history already downloaded for specific workflow id via Run proto. We download history using id and then pass it to replay using this method - https://github.com/temporalio/roadrunner-temporal/blob/feature/replayer-api/rpc.go#L165

wolfy-j: Not sure how to pass id to that replayer.

Sushisource: Oh, yeah, I guess Go doesn't even expose that right now. Don't worry about it then. I think we should remove wf type as a parameter though.

wolfy-j: That’s we can do since it’s already in history.

PR #372, opened 2023-11-16
Scheduling API

### LINKED ISSUE #228: High level API for schedules
EDIT:

More detais here: https://docs.temporal.io/dev-guide/go/features#schedule-a-workflow

roxblnfk:
## What was changed

Added Schedule API.
Marshaller update:
- Better Marshaller toe detection. The type will be detected automatically if there specified only `name` property in Marshal attribute.
- Added converter from Proto to DTO with mapping for common proto Messages. It is used internally for client requests only. We might open it as feature later.
- Added support for OneOf rule for Marshaller.
- Added support for multiple Marshal attributes. It allows to unmarshal value from multiple fields.

## Checklist
1. Closes #228
2. How was this tested: manually; unit tests.

## TODO

- [x] Read API
- [ ] Write API
- [ ] Tests
- [ ] Docs


src/Client/Schedule/Action/StartWorkflowAction.php
```diff
@@ -0,0 +1,287 @@
...
+     * The retry policy for the workflow. Will never exceed {@see self::$workflowExecutionTimeout}.
+     */
+    #[Marshal(name: 'retry_policy')]
+    public readonly RetryOptions $retryPolicy;
+
+    /**
+     * @link https://docs.temporal.io/docs/content/what-is-a-temporal-cron-job/
+     */
+    #[Marshal(name: 'cron_schedule')]
+    public readonly string $cronSchedule;
```
Sushisource, line 94: This shouldn't be in a workflow being started from a schedule

src/Client/Schedule/Info/ScheduleActionResult.php
```diff
@@ -0,0 +1,41 @@
...
+use Temporal\Internal\Marshaller\Meta\Marshal;
+use Temporal\Workflow\WorkflowExecution;
+
+/**
+ * @see \Temporal\Api\Schedule\V1\ScheduleActionResult
+ */
+final class ScheduleActionResult
+{
+    /**
+     * Time that the action was taken (according to the schedule, including jitter).
```
Sushisource, line 18:
```suggestion
     * Time that the action should have been taken (according to the schedule, including jitter).
```

src/Client/Schedule/Policy/ScheduleOverlapPolicy.php
```diff
@@ -0,0 +1,51 @@
...
+ * Controls what happens when a workflow would be started by a schedule, and is already running.
+ *
+ * @see \Temporal\Api\Enums\V1\ScheduleOverlapPolicy
+ */
+enum ScheduleOverlapPolicy: int
+{
+    case Unspecified = 0;
+
+    /**
+     * Means don't start anything. When the workflow completes, the next scheduled event after that
```
Sushisource, line 17: Can eliminate "means" from the start of all these docstrings.

src/Client/Schedule/Policy/SchedulePolicies.php
```diff
@@ -0,0 +1,99 @@
...
+     * time is within this window from the current time.
+     * This value defaults to 60 seconds, and can't be less than 10 seconds.
+     */
+    #[Marshal(name: 'catchup_window', of: Duration::class)]
+    public readonly \DateInterval $catchupWindow;
+
+    /**
+     * If true, and a workflow run fails or times out, turn on "paused".
+     * This applies after retry policies: the full chain of retries must fail to
+     * trigger a pause here.
```
Sushisource, line 43:
```suggestion
     * If true, and a workflow run fails or times out, pause the schedule.
     * This applies after retry policies: the full chain of retries must fail to
     * trigger a pause here.
```

src/Client/Schedule/ScheduleHandle.php
```diff
@@ -0,0 +1,210 @@
...
+    {
+        return $this->id;
+    }
+
+    /**
+     * Update the Schedule.
+     *
+     * NOTE: If two Update calls are made in parallel to the same Schedule there is the potential
+     * for a race condition.
+     */
```
Sushisource, line 58: Probably worth mentioning what the conflict token param is good for, and that it can be obtained from describe

src/Common/TaskQueue/TaskQueue.php
```diff
@@ -0,0 +1,57 @@
...
+namespace Temporal\Common\TaskQueue;
+
+use Temporal\Internal\Marshaller\Meta\Marshal;
+use Temporal\Internal\Traits\CloneWith;
+
+/**
+ * @link https://docs.temporal.io/docs/concepts/task-queues/
+ * @see \Temporal\Api\Taskqueue\V1\TaskQueue
+ */
+final class TaskQueue
```
Sushisource, line 14: I don't think you need this at all. Users never specify `kind` and therefore anywhere that needs to accept a task queue can simply accept a string

src/Common/TaskQueue/TaskQueueKind.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Temporal\Common\TaskQueue;
+
+/**
+ * @see \Temporal\Api\Enums\V1\TaskQueueKind
+ */
+enum TaskQueueKind: int
```
Sushisource, line 10: So this whole thing can probably go too

src/Internal/Marshaller/Mapper/AttributeMapper.php
```diff
@@ -114,20 +114,26 @@ private function getScope(): Scope
      *
      * @param Scope $scope
      *
-     * @return iterable<\ReflectionProperty, MarshallingRule|null>
+     * @return iterable<\ReflectionProperty, array{MarshallingRule|null, bool}>
      */
     private function getPropertyMappings(Scope $scope): iterable
     {
+        // $stack = new SplStack();
```
Sushisource, line 121: Leftover comment?

Sushisource (commented):
Generally looks good to me, just a few things to change, but I'm approving now to unblock.

I didn't review the marshaling changes in great detail, mostly focused on the user-facing stuff.

Sushisource (approved):
Forgot to actually approve

src/Common/TaskQueue/TaskQueue.php
roxblnfk: I've removed the `TaskQueue::withKind()` method, leaving only `TaskQueue::withName()`. It will be difficult to get rid of this DTO because of the automatic mapping. However, in methods that take `TaskQueue` from the user, it's possible to specify a `string`, not only a `TaskQueue` object: 

```php
// class StartWorkflowAction
public function withTaskQueue(string|TaskQueue $taskQueue): self
```

roxblnfk:
It seems I forgot to create a DTO for the patch() method 🤦

Issue #272, opened 2023-01-04
[Bug] PHP 8 named arguments don't work


roxblnfk:
### What are you really trying to do?

I am trying to call workflow handler using PHP8 named arguments.

```php
// Workflow handker signature:
public function handler(bool $foo = false, bool $bar = false): iterable;

// Calling
$workflowStub->handler(bar: true);
```
In this case the handler will be started with `[true, false]` values

### Environment/Versions

PHP 8.1, SDK-PHP 2.3.2

wolfy-j:
Named arguments are not allowed due to limited language support.

roxblnfk:
> Named arguments are not allowed due to limited language support.

In the SDK-PHP we support PHP syntax. When named arguments are used we will send all skipped default values:

```php
// activity method
public function doSomething(?int $number = 0, string $foo = 'default', ?Bar $bar = null);
```

If we call `doSomething(foo: 'custom');` values `[0, 'custom']` will be sent (without names of course)

wolfy-j:
Ah... ok, so it's on packing side.

PR #376, opened 2023-12-03
Pass named functions arguments in correct order and with default values

shanginn:
## What was changed
Allow for named arguments in workflow handler, signals and activities params

## Why?
Because it's possible in PHP

## Checklist

1. Closes [Issue 282](https://github.com/temporalio/sdk-php/issues/272)

2. Tested in NamedArgumentsTestCase


roxblnfk:
Hello. Thank you for your contribution to the Temporal PHP SDK.
I've done an initial review and made some edits. However, there's still a lot of work to be done. Please note that there is an opportunity to create different stubs (like `ChildWorkflow`) through the `Workflow::` facade. We need to ensure that named arguments also work in such a flow.

roxblnfk:
Need to fix arguments sorter. See the new unit tests

roxblnfk (approved):
LGTM
@shanginn need to fix psalm issues

PR #377, opened 2023-12-05
Fix Schedule API, cleaning and maintenance

roxblnfk:
## What was changed
- Removed `buildkite`.
- Added Symfony 7 packages.
- Fixed bugs in Schedule API.

## Why?
- Buildkite isn't used in this package.
- New versions of Symfony packages need to be added for compatibility with all frameworks.
- Bugs in Schedule were identified during more in-depth manual testing of the SDK.

## Checklist

1. Part of #228


roxblnfk:
@Sushisource I think it need to disable buildkite integration with this package
image

Sushisource (approved):
Turned off the buildkite trigger

PR #391, opened 2024-01-17
Optimize garbage collecting

roxblnfk:
## What was changed

- Added `destroy()` methods to Workflow-related classes
- Added a `GarbageCollector` which is responsible for the frequency of calling the finding circular references function (garbage collecting).
- `phpunit` has been updated to version 10. This is a necessary measure because changes in the PR broke some tests that couldn't be handled using `phpunit` 9.

## Why?

A few months ago, when we were troubleshooting a memory leak issue, I added a call to the `gc_collect_cycles()` method after each `DestroyWorkflow` command execution.

Garbage collection is quite a costly operation, and Workflows in Temporal are created and destroyed very frequently. We can verify this by looking at the [metrics provided by the user Pyjac](https://app.slack.com/client/TNWA8QCGZ/C01LK9FAMM0/thread/C01LK9FAMM0-1700254848.292919?cdn_fallback=1)

image

We recently returned to the issue of potential memory leaks. In the process, some important optimizations were implemented:
1. Added methods for manual destruction of Workflow-related objects, which free resources and break circular references between objects. Thus, all SDK code works without leaks, using PHP's self-cleaning mechanisms.
2. We left the call to `gc_collect_cycles()`, but introduced control over the frequency of these calls. The collecting of circular references is now called by a 30-second timeout or when 1000 Workflows are destroyed. This way we are combating memory leaks in user code.

After this update, we are seeing a significant increase in the throughput of the Workflow worker and almost no TaskTimeout errors under highload.

<details>

Before

Screenshot 2024-01-16 153241

After
Screenshot 2024-01-16 161004

</details>

### DIFF
--- a/src/Internal/Transport/Router/DestroyWorkflow.php
+++ b/src/Internal/Transport/Router/DestroyWorkflow.php
@@ -27,7 +27,9 @@ class DestroyWorkflow extends WorkflowProcessAwareRoute
      */
     private const ERROR_PROCESS_NOT_DEFINED = 'Unable to kill workflow because workflow process #%s was not found';
 
+    /** Maximum number of ticks before GC call. */
     private const GC_THRESHOLD = 1000;
+    /** Interval between GC calls in seconds. */
     private const GC_TIMEOUT_SECONDS = 30;
 
     private GarbageCollector $gc;
@@ -64,6 +66,8 @@ class DestroyWorkflow extends WorkflowProcessAwareRoute
             'finally',
             function () use ($process) {
                 $process->destroy();
+
+                // Collect garbage if needed
                 if ($this->gc->check()) {
                     $this->gc->collect();
                 }
--- a/src/Internal/Transport/Router/DestroyWorkflow.php
+++ b/src/Internal/Transport/Router/DestroyWorkflow.php
@@ -14,19 +14,14 @@ namespace Temporal\Internal\Transport\Router;
 use React\Promise\Deferred;
 use Temporal\DataConverter\EncodedValues;
 use Temporal\Exception\DestructMemorizedInstanceException;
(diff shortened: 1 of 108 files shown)


src/Internal/Transport/Router/DestroyWorkflow.php
```diff
@@ -43,13 +59,20 @@ public function handle(ServerRequestInterface $request, array $headers, Deferred
...
-        $process = $this->running->find($runId);
-        if ($process === null) {
-            throw new \InvalidArgumentException(\sprintf(self::ERROR_PROCESS_NOT_DEFINED, $runId));
-        }
+        $process = $this->running->pull($runId);
 
-        $this->running->pull($runId);
         $process->cancel(new DestructMemorizedInstanceException());
+        $this->loop->once(
+            'finally',
```
wolfy-j, line 66: should it be using ON_FINALLY constant?

wolfy-j: do we really need it?

src/Internal/Workflow/Process/Scope.php
```diff
@@ -245,8 +246,8 @@ public function cancel(\Throwable $reason = null): void
 
         foreach ($this->onCancel as $i => $handler) {
             $this->makeCurrent();
-            $handler($reason);
             unset($this->onCancel[$i]);
+            $handler($reason);
```
wolfy-j, line 250: why?

wolfy-j (approved):

src/Internal/Workflow/Process/Scope.php
roxblnfk: I'm not sure if there's a way to call cancel inside cancel through this callback, but I prefer to deduplicate the function call without waiting for it to happen.

PR #406, opened 2024-03-11
Fix FailureConverter when an exception CallStack contains an incomplete set of keys

actuallymab:
## What was changed
All elements returned by https://www.php.net/manual/en/function.debug-backtrace.php are possible, but not guaranteed. Safeguard added for elements without "class" property.

## Why?
I encountered this problem when I was using https://laravel.com/docs/10.x/helpers#method-throw-if helper function, that doesn't belong to a class. See screenshot:

Screenshot 2024-03-11 at 12 07 15

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -395,9 +395,6 @@
     <InvalidOperand>
       <code><![CDATA[count($arg)]]></code>
     </InvalidOperand>
-    <MissingClosureReturnType>
-      <code><![CDATA[static fn() => \sprintf(]]></code>
-    </MissingClosureReturnType>
     <NullableReturnStatement>
       <code><![CDATA[$e->getFailure()]]></code>
     </NullableReturnStatement>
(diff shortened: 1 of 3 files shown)


roxblnfk (approved):

roxblnfk:
Hello.
Is it possible to add a test there?
And you have to sign CLA

actuallymab:
> Hello. Is it possible to add a test there? And you have to sign CLA

thanks for your quick response on this PR @roxblnfk 

it seems like there is no test covering this piece yet, do you want me to introduce new tests under `\Temporal\Tests\Unit\Exception\FailureConverterTestCase`?

roxblnfk:
Yes, it would be great to add a few Unit tests. I intend to release a fix as soon as possible, and I will also add something to improve the code understanding by the static analyzer.

actuallymab:
@roxblnfk thanks for your psalm changes, I've added two unit tests which aims to cover the change. Please let me know if you have any feedback.

roxblnfk:
@actuallymab thank you for your contribution. The fix is now available in version [2.8.1](https://github.com/temporalio/sdk-php/releases/tag/v2.8.1).

PR #407, opened 2024-03-20
Add generation of API documentation

msmakouz:


.github/workflows/docs.yml
```diff
@@ -0,0 +1,35 @@
...
+        id: phpdocumentor-cache
+        uses: actions/cache@v3
+        with:
+          path: .phpdoc/cache
+          key: ${{ runner.os }}-phpdocumentor-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-phpdocumentor-
+
+      - name: Build with documentation
+        run: docker run --rm --volume "$(pwd):/data" phpdoc/phpdoc:3.4 -vv --template default
```
roxblnfk, line 32: I think we can use `latest` or `3` there

roxblnfk (approved):
- [ ]

PR #412, opened 2024-04-03
fix: duration json type parsing fix

Zylius:
## What was changed
Added ability to parse DurationJsonType instance.
## Why?
It no longer works, and if you need to pass it around to other services or save ActivityOptions somewhere, stuff breaks. I've made this as a check, so it should be BC, and messages using older format shouldn't fail during deployment.

### DIFF
--- a/src/Internal/Marshaller/Type/DurationJsonType.php
+++ b/src/Internal/Marshaller/Type/DurationJsonType.php
@@ -70,6 +70,11 @@ class DurationJsonType extends Type implements DetectableTypeInterface, RuleFact
      */
     public function parse($value, $current): CarbonInterval
     {
+        if (is_array($value) && isset($value['seconds']) && isset($value['nanos'])) {
+            // The highest precision is milliseconds either way.
+            $value = ($value['seconds'] * 1000000000 + $value['nanos']) / 1000;
+        }
+
         return DateInterval::parse($value);
     }
 }
--- a/src/Internal/Marshaller/Type/DurationJsonType.php
+++ b/src/Internal/Marshaller/Type/DurationJsonType.php
@@ -31,7 +31,7 @@ class DurationJsonType extends Type implements DetectableTypeInterface, RuleFact
 
     /**
      * @param MarshallerInterface $marshaller
-     * @param string $format Fall back format for parsing when the value is not an array.
+     * @param DateIntervalFormat $format Fall back format for parsing when the value is not an array.
      */
     public function __construct(MarshallerInterface $marshaller, string $format = DateInterval::FORMAT_NANOSECONDS)
     {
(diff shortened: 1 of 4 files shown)


Zylius:
Any ideas when this could be merged? :pleading_face:

roxblnfk:
Could you add a test case?

Zylius:
I cannot offer you tests, as they're functional and integrated within our system, but this snippet is basically already a test, just needs an assertion.

Without the code updated this will throw InvalidArgumentException

```php
$type = new \Temporal\Internal\Marshaller\Type\DurationJsonType(new class implements \Temporal\Internal\Marshaller\MarshallerInterface
{
    public function marshal(object $from): mixed
    {
        return null;
    }

    public function unmarshal(array $from, object $to): object
    {
        return new \stdClass();
    }
});

$type->parse($type->serialize(1), null);
```

Zylius:
Got a few free minutes, added a test :) without the fix it fails

roxblnfk:
Do you need the ability to unmarshal ActivityOptions from the old JSON structure?

Zylius:
You saying where are more changes? I think only intervals changed

Here's the JSON payload exmple from 2.7.*
```json
{
    "TaskQueueName": "test_queue",
    "ScheduleToCloseTimeout": 0,
    "ScheduleToStartTimeout": 0,
    "StartToCloseTimeout": 5000000000,
    "HeartbeatTimeout": 0,
    "WaitForCancellation": false,
    "ActivityID": "",
    "RetryPolicy": {
        "initial_interval": 0,
        "backoff_coefficient": 1,
        "maximum_interval": 0,
        "maximum_attempts": 4,
        "non_retryable_error_types": []
    }
}
```

Here is 2.8
```json
{
    "TaskQueueName": "test_queue",
    "ScheduleToCloseTimeout": 0,
    "ScheduleToStartTimeout": 0,
    "StartToCloseTimeout": 5000000000,
    "HeartbeatTimeout": 0,
    "WaitForCancellation": false,
    "ActivityID": "",
    "RetryPolicy": {
        "initial_interval": {
            "seconds": 0,
            "nanos": 0
        },
        "backoff_coefficient": 1,
        "maximum_interval": {
            "seconds": 0,
            "nanos": 0
        },
        "maximum_attempts": 4,
        "non_retryable_error_types": []
    }
}
```

roxblnfk:
It should be nanoseconds, right?
```
        "initial_interval": 0,
        "maximum_interval": 0,
```

Zylius:
> It should be nanoseconds, right?
> 
> ```
>         "initial_interval": 0,
>         "maximum_interval": 0,
> ```

Yes the old format is in nanoseconds

roxblnfk (approved):

PR #423, opened 2024-04-17
[BUG] Broken DateInterval to protobuf Duration casting

Zylius:
## What was changed
Fixed DateInterval to protobuf duration casting.

## Why?
Would cause ```proto_codec_parse_message: json: cannot unmarshal number 5000000000 into Go struct field Duration.Nanos of type int32``` in some cases, specifically:

```
DurationJsonType:93 turns from protobuf format to CarbonInterval
DateInterval:117 division causes f property (float) in CarbonInterval to be set
DateInterval:165 set f propoerty in CarbonInterval causes an interval of 5 seconds to be parsed into protobuf object of [seconds => 5, nanos => 5 000 000 000]

5 000 000 000 doesnt fit into int32 of Duration struct and roadrunner/temporal panics

This would always be an issue with carbon 2 or 3, as the f always can include seconds
```
## Checklist

How was this tested:
Unit tests


roxblnfk (approved):

roxblnfk:
Oh, sorry.🙀 Could you reopen the PR from `2.8.x` into `2.8.x`?

Zylius:
Ohh thank you I was afk, I see it's all resolved now

PR #378, opened 2023-12-13
[Maintenance] Add tests for Schedule API, polishing, fixes

roxblnfk:
## What was changed

Added tests for Schedule API.
Added `ScheduleClientInterface`.
Usability has been improved.

Enum `WorkflowIdReusePolicy` has removed because there is the same in the `common` directory (class `IdReusePolicy` with constants). The `IdReusePolicy` has been converted into PHP enum with constants.

`ServiceClient::withInterceptorsPipeline()` has been renamed into `withInterceptorPipeline()`

The methods `ServcieClient::getServerCapabilities()` and `ServcieClient::setServerCapabilities()` have been added to the auto-generated interface.
Now, they will not be overwritten when the interface is regenerated.

---

The minimum version threshold for Symfony packages has been raised to 5.4. This is the lowest supported LTS version.
The 4.4 version ceased to be supported in November 2023.
image

---

Fixex `HistoryLength` update when a Signal method is called

---

Updated psalm baseline

## Checklist

Closes #228

PR #427, opened 2024-04-30
fix: bring back psr/log ^1

Zylius:
## What was changed
Added support for psr/log ^1

## Why?
It makes now sense to support ^2 |^3 but not support ^1, the changes are only in the way types are specified in interfaces.  Symfony and projects like that all support ^1|^2|^3.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "google/protobuf": "^3.22",
         "grpc/grpc": "^1.42",
         "nesbot/carbon": "^2.66 || ^3.0.2",
-        "psr/log": "^2.0 || ^3.0",
+        "psr/log": "^1.0 || ^2.0 || ^3.0",
         "ramsey/uuid": "^4.7",
         "react/promise": "^2.9",
         "roadrunner-php/roadrunner-api-dto": "^1.5.0 <1.7.0",


Zylius:
I can see that the "lowest-version" check will fail but I think this has negative impact on usability of this project. For example it cannot be used with symfony 5.3 and laminas-di 3.4

roxblnfk:
> For example it cannot be used with symfony 5.3 and laminas-di 3.4

Hi 👋
Yes, we have upgraded the minimum Symfony version from 4.4 to 5.4.

> _https://github.com/temporalio/sdk-php/pull/378_
> The minimum version threshold for Symfony packages has been raised to 5.4. This is the lowest supported LTS version.
> The 4.4 version ceased to be supported in November 2023.
> image

Zylius:
> `composer why-not psr/log 1.0.0`
> 
> >

Yeah, The 5.4 or 5.3 isn't the issue, it doesn't matter, symfony supports psr/log ^3 from symfony 6 only. 

As for the why-not, all the packages listed as why-not are spiral php packages which all have the "validate-prefer-lowest" for psr/log, which forces outside projects to update the psr/log. Which sometimes cannot be done due to other external packages.

So the public usability of the sdk decreases for no logical or practical reason.

Symfony 6/7 all support psr/log "^1|^2|^3"

I would encourage you guys to rethink the "validate-prefer-lowest" check, it's good for internal packages probably, but not so for public ones, in this specific case at least.

Zylius:
Yees, thank you

PR #467, opened 2024-07-01
Fix EncodedValues to return null if possible when there is no value in Payloads

roxblnfk:
## What was changed

Added more cases for https://github.com/temporalio/sdk-php/pull/442

### DIFF
--- a/tests/Unit/Protocol/EncodingTestCase.php
+++ b/tests/Unit/Protocol/EncodingTestCase.php
@@ -11,9 +11,12 @@ declare(strict_types=1);
 
 namespace Temporal\Tests\Unit\Protocol;
 
+use PHPUnit\Framework\Attributes\DataProvider;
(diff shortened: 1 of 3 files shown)


wolfy-j (approved):

tests/Unit/Protocol/EncodingTestCase.php
```diff
@@ -27,4 +30,77 @@ public function nullValuesAreReturned(): void
...
+    {
+        $encodedValues = EncodedValues::fromPayloadCollection(new \ArrayIterator([]));
+
+        self::expectException(\LogicException::class);
+        self::expectExceptionMessage('DataConverter is not set');
+
+        $encodedValues->getValue(0, $type);
+    }
+
+    private static function getRturnType(\Closure $closure): \ReflectionType
```
igancev, line 102: Misprint in method name

PR #487, opened 2024-08-15
Remove final from WorkflowInterface

Nyholm:
## What was changed

I made the WorkflowInterface attribute not final

## Why?

I am extending this attribute to allow myself to add more properties. The additional properties will help me with building the container and workers. 

Example:

```php
use Temporal\Workflow\WorkflowInterface;

class MyWorkflowInterface extends WorkflowInterface
{
    public function __construct(
        public string $workerName, 
    ){}
}
```

Now I can create a bunch of worker configuration and map it with my workflows and let my applications service container build the workers for me. 

This is similar to `ActivityInterface` attribute. 

## Checklist
1. How was this tested:

I did not test at all. 

2. Any docs updates needed?

Nope

### DIFF
--- a/src/Workflow/WorkflowInterface.php
+++ b/src/Workflow/WorkflowInterface.php
@@ -26,6 +26,6 @@ use Spiral\Attributes\NamedArgumentConstructor;
  *  Problem is relevant for doctrine/annotations 1.11 or lower on any PHP version.
  */
 #[\Attribute(\Attribute::TARGET_CLASS), NamedArgumentConstructor]
-final class WorkflowInterface
+class WorkflowInterface
 {
 }


roxblnfk:
Hello 👋
Have you considered creating a separate attribute like other integrations do?
For example, in integration with Spiral, [this attribute](https://github.com/spiral/temporal-bridge/blob/3.x/src/Attribute/AssignWorker.php) is used to specify the TaskQueue name when registering workers.

roxblnfk (approved):

Nyholm:
Hello :)

Yes. That is what I do currently. But that forces my user to have two attributes. 

But that is not the end of the world. They have one attribute on the workflow interface and one on the workflow implementation.

butschster:
> Hello :)
> 
> Yes. That is what I do currently. But that forces my user to have two attributes.
> 
> But that is not the end of the world. They have one attribute on the workflow interface and one on the workflow implementation.

When working with Temporal workflows that need to be shared across different services, we can use interfaces in a shared package. This approach allows services that use this package to invoke the workflow. Here's how it's implemented:

1. Interface in a shared package:

```php
// Shared package
<?php
declare(strict_types=1);
namespace Internal\Shared\Temporal\Workflow\Subtitle;
use GRPC\Services\Subtitles\v1\Request\GenerateSubtitleRequest;
use GRPC\Services\Subtitles\v1\Subtitle;
use Internal\Shared\Temporal\TaskQueue;
use Spiral\TemporalBridge\Attribute\AssignWorker;
use Temporal\Support\VirtualPromise;
use Temporal\Workflow\WorkflowInterface;
use Temporal\Workflow\WorkflowMethod;

#[AssignWorker(TaskQueue::SUBTITLE_GENERATOR)]
#[WorkflowInterface]
interface GenerateSubtitleWorkflow
{
    /**
     * @return VirtualPromise<Subtitle>
     */
    #[WorkflowMethod]
    public function generate(GenerateSubtitleRequest $request);
}
```

2. Implementation in a service that handles the workflow:

```php
<?php
declare(strict_types=1);
namespace App\Endpoint\Temporal;
use App\Endpoint\Temporal\Subtitle\SubtitleServiceStub;
use GRPC\Services\Subtitles\v1\Request\CreateRequest;
use GRPC\Services\Subtitles\v1\Request\GenerateSubtitleRequest;
use GRPC\Services\Subtitles\v1\Request\SubtitleProcessed;
use GRPC\Services\Subtitles\v1\Request\TranscribeRequest;
use GRPC\Services\Subtitles\v1\SubtitleRecord;
use Internal\Shared\Temporal\Workflow\Audio\TranscribeAudioWorkflowStub;
use Internal\Shared\Temporal\Workflow\Subtitle\GenerateSubtitleWorkflow;
use Temporal\Internal\Workflow\ActivityProxy;

final class ProcessAudioWorkflow implements GenerateSubtitleWorkflow
{
    private ActivityProxy|Subtitle\SubtitleActivity $subtitles;

    public function __construct()
    {
        $this->subtitles = SubtitleServiceStub::create();
    }

    public function generate(GenerateSubtitleRequest $request)
    {
        // 1. Check if audio is mp3
        // 2. Check if subtitles exist for given audio hash
        // - If so, return UUID
        // 3. Create subtitle record with status pending
        // 4. Start transcribe
        // Set UUID to subtitle
        // - persist to database
        return $subtitle;
    }
}
```

This setup uses the Spiral framework to handle Temporal. When the service starts, the framework looks for all classes or interfaces that have the `WorkflowInterface` attribute. If the attribute is on an interface and it has an implementation inside the service, it registers it in the Temporal server.

butschster:
Oh) Your idea more about workers. I think, it;s better to use separate attribute, like `AssignWorker` in my example.In this case you don't need to extend final attributes)

wolfy-j (approved):

roxblnfk:
Will be available with 2.11.0.

Nyholm:
Thank you for merging

PR #488, opened 2024-08-16
Warn about unfinished handlers

### LINKED ISSUE #446: Warn or error when update handlers dangle across CAN or workflow exit
We can recommend the await-all-handlers or drop-all-handlers methods. TBD whether error or warning.

roxblnfk:
## What was changed

Added logging of unfinished Signal and Update handlers when a Workflow finishes.

> [!WARNING]
> Warn will happen in cases of Workflow **success**, **cancel**, **continue-as-new**, **fail** with retryable, **fail** with non-retryable.

If there are unfinished handlers of both types, two warnings will be sent.

To disable such warnings a new feature flag `$warnOnWorkflowUnfinishedHandlers` has been added:
```php
\Temporal\FeatureFlags::$warnOnWorkflowUnfinishedHandlers = false;
```

Examples:

> Workflow `Extra_Workflow_AllHandlersFinished` cancelled while update handlers are still running. This may have interrupted work that the update handler was doing, and the client that sent the update will receive a 'workflow execution already completed' RPCError instead of the update result. You can wait for all update and signal handlers to complete by using `yield Workflow::await(Workflow::allHandlersFinished(...));`. Alternatively, if both you and the clients sending the update are okay with interrupting running handlers when the workflow finishes, and causing clients to receive errors, then you can disable this warning via the update handler attribute: `#[UpdateMethod(unfinishedPolicy: HandlerUnfinishedPolicy::Abandon)]`. The following updates were unfinished (and warnings were not disabled for their handler): `await` id:5e51ad4d-041c-4604-8b0a-cda641169da6, `await` id:1ea22590-aafa-4d27-aea4-9f0b7122b33d, `await` id:12059abd-2097-452b-b4de-36978c1c45da, `await` id:ca26f433-0974-4332-bc01-91c32e97dcc0, `await` id:ec827b4a-801c-453f-864c-b042bbcda3ea, `await` id:4b4dcf64-043c-4de2-b1fa-7431aedabf4c, `await` id:57839a31-34db-4560-b10a-4afc083133cc, `await` id:14158ca5-2d65-47f4-824c-7e15705219b5

> Workflow `Extra_Workflow_AllHandlersFinished` finished while signal handlers are still running. This may have interrupted work that the signal handler was doing. You can wait for all update and signal handlers to complete by using `yield Workflow::await(Workflow::allHandlersFinished(...));`. Alternatively, if both you and the clients sending the signal are okay with interrupting running handlers when the workflow finishes, and causing clients to receive errors, then you can disable this warning via the signal handler attribute: `#SignalMethod(unfinishedPolicy: HandlerUnfinishedPolicy::Abandon)]`. The following signals were unfinished (and warnings were not disabled for their handler): `resolve` x2, `await` x8

## Checklist

1. Closes #446
3. Any docs updates needed?

### DIFF
--- a/tests/Acceptance/Extra/Workflow/AllHandlersFinishedTest.php
+++ b/tests/Acceptance/Extra/Workflow/AllHandlersFinishedTest.php
@@ -128,17 +128,61 @@ class AllHandlersFinishedTest extends TestCase
             'Workflow result contains resolved values',
         );
     }
+
+    #[Test]
+    public function warnUnfinishedSignals(
+        #[Stub('Extra_Workflow_AllHandlersFinished')] WorkflowStubInterface $stub,
+    ): void {
+        $this->markTestSkipped("Can't check the log yet");
+
+        for ($i = 0; $i < 8; $i++) {
+            /** @see TestWorkflow::addFromSignal() */
+            $stub->signal('await', "key-$i");
+        }
+        /** @see TestWorkflow::resolveFromSignal() */
+        $stub->signal('resolve', 'foo');
+
+        // Finish the workflow
+        $stub->signal('exit');
+        $stub->getResult(timeout: 1);
+
+        // todo Check that `await` signal with count was mentioned in the logs
+    }
+
+    #[Test]
+    public function warnUnfinishedUpdates(
+        #[Stub('Extra_Workflow_AllHandlersFinished')] WorkflowStubInterface $stub,
+    ): void {
+        $this->markTestSkipped("Can't check the log yet");
+
+        for ($i = 0; $i < 8; $i++) {
+            /** @see TestWorkflow::addFromSignal() */
+            $stub->startUpdate('await', "key-$i");
+        }
+        /** @see TestWorkflow::resolveFromSignal() */
+        $stub->startUpdate('resolve', 'foo');
+
+        // Finish the workflow
+        $stub->signal('exit');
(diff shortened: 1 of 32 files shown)


wolfy-j (approved):

tests/Acceptance/Extra/Workflow/AllHandlersFinishedTest.php
```diff
@@ -128,17 +131,88 @@ public function signalHandlersWithManyCalls(
...
+        for ($i = 0; $i < 8; $i++) {
+            /** @see TestWorkflow::addFromSignal() */
+            $stub->signal('await', "key-$i");
+        }
+
+        // Finish the workflow
+        $stub->signal('exit');
+        $stub->getResult(timeout: 1);
+
+        // todo Check that `await` signal with count was mentioned in the logs
```
Sushisource, line 154: Looks like a bit of leftover? Below as well

Sushisource (approved):
This looks really nice. Makes sense to me. Thanks guys!

tests/Acceptance/Extra/Workflow/AllHandlersFinishedTest.php
roxblnfk: We currently don't have the capability to test this, so these TODOs will likely remain for the time being

PR #492, opened 2024-08-24
Add getter for ActivityPrototype::$factory

Nyholm:
## What was changed
Adding a getter for ActivityPrototype::$factory

## Why?

Whenever I start my workers, I replace all `ActivityPrototype` with a decorated version of my activities. I do this to add support for EventDispatcher. Ie I dispatch event before and after an activity is executed. 

So I need to rebuild the `ActivityPrototype`. 

## Checklist

2. How was this tested:
Unit tests

3. Any docs updates needed?
Nope

### DIFF
--- a/src/Internal/Declaration/Prototype/ActivityPrototype.php
+++ b/src/Internal/Declaration/Prototype/ActivityPrototype.php
@@ -112,4 +112,9 @@ final class ActivityPrototype extends Prototype
     {
         return $this->isLocalActivity;
     }
+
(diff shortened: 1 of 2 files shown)


wolfy-j (approved):

roxblnfk (approved):
Hi @Nyholm 👋
I see no obstacles to merging the PR, however note that you are replacing internal classes. You can intercept calls and run Activities in a legal way through interceptors.

PR #494, opened 2024-08-24
Remove FeatureFlags pattern

Nyholm:
I see that https://github.com/temporalio/sdk-php/pull/483 introduce FeatureFlag class. That is a pattern that could be successfully used in an application but should be avoided in a shared library. 

It is basically shared global memory. Ie it is as bad as using `$_SERVER` to store data in. Im not sure if this specific is bad/dangerous in any way, but I think we should avoid open this door. 

## What was changed
Removed class `FeatureFlag` and use proper constructor arguments instead

## Why?

Shared memory is bad.


Nyholm:
Could you please run the CI? It will make it easier for me to have this PR "ready to be merged".

roxblnfk:
Sorry for the delay. I partially agree with your remark.

As you might have guessed, these flags change the behavior of the SDK at the level of the entire PHP worker (process). They are introduced for a consistent migration to more correct behavior, which will be established in the next major version.

In my opinion, passing individual flags through the constructor might be a mistake because:
- Some experimental flags may be removed between minor versions.
- The behavior should not differ for different workers. Here, a global state looks better.

The solution that might look best is to create a common builder (or two: one for the client and one for the worker), which is configured once and from which any configured class (WorkflowClient, ScheduleClient, ServiceClient, cloud API client, WorkerFactory, etc.) can be obtained without duplicating internal services. In this case, the DTO with flags will be among the other common services.

This also would help to resolve some inconveniences when using multiple types of clients that need to use the same configs and services (marshaller, encoders, credentials, etc.).
Passing flags to such a builder would be preferable to what we have now.

The release is scheduled for September, and I'm not sure I will have enough time to add the Builder. However, upon adding such a builder, the FeatureFlags class can be easily deprecated.

roxblnfk:
@Nyholm hi 👋
Could you please leave feedback on this PR? https://github.com/temporalio/sdk-php/pull/505

roxblnfk:
Closed in favor of #510

PR #496, opened 2024-08-25
Make sure we run php-cs-fixer as part of our CI

Nyholm:
## What was changed
I see that we have php-cs-fixer config and library installed, but we dont run it as part of the CI. 

## Why?
Code style rules should be forced and handled automatically. This should not take up time for the reviewer or the contributors. 

## Checklist

Please review this PR by first looking at my manual changes in f45a9e2cd613619587a3cce93df4c7b435c5777b, then looking at the automatic changes in 85e48c3534d96176d896e20754a9464cecb47047.

### DIFF
--- a/.github/workflows/code-style.yml
+++ b/.github/workflows/code-style.yml
@@ -44,7 +44,7 @@ jobs:
 
   php-cs-fixer:
     name: PHP-CS-Fixer
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
 
     steps:
       - name: Checkout code
@@ -56,6 +56,9 @@ jobs:
           php-version: 8.1
           coverage: none
           tools: cs2pr
+
+      - name: Download dependencies
+        uses: ramsey/composer-install@v3
 
       - name: Display PHP-CS-Fixer version
         run: php-cs-fixer --version
--- a/.github/workflows/code-style.yml
+++ b/.github/workflows/code-style.yml
@@ -21,26 +21,12 @@ jobs:
 
       - name: Check Out Code
         uses: actions/checkout@v4
-        with:
-          fetch-depth: 1
 
-      - name: Get Composer Cache Directory
-        id: composer-cache
-        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
-
-      - name: Cache Dependencies
-        uses: actions/cache@v3
-        with:
-          path: ${{ steps.composer-cache.outputs.dir }}
-          key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
-          restore-keys: php-${{ matrix.php }}-${{ runner.os }}-composer-
-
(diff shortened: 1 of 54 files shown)


Nyholm:
Im not a big fan when new contributors create a PR with code style changes. But I think this is okey because the code style rules are already defined.

.github/workflows/code-style.yml
```diff
@@ -21,22 +21,33 @@ jobs:
 
       - name: Check Out Code
         uses: actions/checkout@v4
-        with:
-          fetch-depth: 1
```
Nyholm, line 25: Version 4 defaults to `fetch-depth: 1`

See https://github.com/actions/checkout?tab=readme-ov-file#usage

.github/workflows/code-style.yml
```diff
@@ -21,22 +21,33 @@ jobs:
...
 
-      - name: Get Composer Cache Directory
-        id: composer-cache
-        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
```
Nyholm, line 29: `ramsey/composer-install@v3` will handle the cache automatically. 

See https://github.com/ramsey/composer-install?tab=readme-ov-file#about

.github/workflows/code-style.yml
```diff
@@ -21,22 +21,33 @@ jobs:
...
+      - name: Download dependencies
+        uses: ramsey/composer-install@v3
+
+      - name: Run Tests
+        run: vendor/bin/psalm --output-format=github
```
Nyholm, line 29: With `--output-format=github` it will make github review comments on the issues. 

(This may be detected automatically, but I could not find a reference for it in the docs)

roxblnfk (commented):
Hi 👋, I very appreciate your effort to improve the SDK codebase. However, it is better to discuss changes in a issue beforehand.

Reasons I cannot accept this PR right now:
- The existing CS config is outdated and does not meet expectations.
- Sanitizing the entire code should be done when it does not conflict with other working branches.

Let's leave the CS question in an issue for the next sprint when no features will be in progress.

I'll prepare GitHub Actions with up-to-date configs and request a bot for automatic changes pushing.

Nyholm:
I've updated the PR to be more restrictive in the changes it is doing. The current CI fixes are very easy to handle in conflicts..
 Also, there are CS fixes, just "accept yours" and you are good. 

I understand this is a PR that should be made by maintainers. I did not look properly, but I didn't find the discussion/argument to not run CS. I assumed it was something forgotten. 

I'll leave this PR now, if maintainers want to take parts of it later or want me to keep it updated, just let me know.

roxblnfk:
Hi. In [this PR](https://github.com/temporalio/sdk-php/pull/506) I updated the CS Fixer rules and added a Workflow in CI that will automatically fix CS issues. Thank you for bringing this to my attention.

PR #505, opened 2024-09-17
Move static flags into DTO for Worker Factory

roxblnfk:
## What was changed

Static worker flags have been refactored to DTOs.
Now, the DTO with flags can be passed to the service container when creating the Worker Factory.

```php
$factory = WorkerFactory::create(
    converter: $converter,
    flags: FeatureFlags::createDefaults(),
);
```

`FeatureFlags` class has two factories: `createDefaults` and `createLegacy` (any ideas for better naming?).
The first sets the flags to meet current SDK requirements.
`createLegacy` sets the flags to ensure compatibility with the project codebase written for SDK versions `<2.11.0`.

By default we create FeatureFlags using legacy set

## Why?

See https://github.com/temporalio/sdk-php/pull/494#issuecomment-2311005196


roxblnfk:
Closed in favor of #510

Issue #501, opened 2024-09-10
Repair tests in CI


roxblnfk:
Sometimes Acceptance and Functional tests in CI end with error code -1073741819.
However, the tests themselves pass completely.
It is necessary to identify the source of the problem. This may be related to how RoadRunner or Temporal CLI are terminated.

![image](https://github.com/user-attachments/assets/62ca958e-ee03-412a-bd95-40fcb9805719)

roxblnfk:
Reproduced on the next env:
- Windows
- PHP 8.3.x
- without protobuf extension (with it works well)

roxblnfk:
Feature and Acceptance tests on Windows still fail with error code `-1073741819`, even though the tests themselves pass and ext-protobuf is installed.

PR #520, opened 2024-10-27
Use environment options to create RRActivityInvocationCache

jmortlock:
## What was changed
For the ActivityMocker invocation cache a hard-coded address was used for the roadrunner RPC settings, in certain scenarios it is desirable to change it. This is inline with other parts of the code that set the RPC address.

## Why?
On my local PC i am running temporal/roadrunner combination for local development purposes
For PHPUnit tests i am spinning up the temporal-test-server (for the time shifting ability) and another roadrunner which connects to this temporal server. The RPC settings on the second roadrunner need to be different to not clash with the first process.

## Checklist

1. Closes 

2. How was this tested:
Tested by making changes locally

3. Any docs updates needed?
None

### DIFF
--- a/src/Worker/ActivityInvocationCache/RoadRunnerActivityInvocationCache.php
+++ b/src/Worker/ActivityInvocationCache/RoadRunnerActivityInvocationCache.php
@@ -6,6 +6,7 @@ namespace Temporal\Worker\ActivityInvocationCache;
 
 use React\Promise\PromiseInterface;
 use Spiral\Goridge\RPC\RPC;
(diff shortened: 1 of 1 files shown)


jmortlock:
Not sure if the test failures are related to the change? in theory the environment returns the same value if the envvar is not set so there should be no real change in behaviour.

roxblnfk:
> Not sure if the test failures are related to the change?

This is not related to the changes 👌  The tests pass successfully but sometimes end with a wrong code. It's something Windows-specific that still needs to be investigated #501

PR #526, opened 2024-11-13
Add `WorkflowUpdateResultException`

roxblnfk:
## What was changed

Added the `WorkflowUpdateResultException`, which is thrown by `UpdateHandle->getResult()` when a Workflow Update Request is stuck in the ADMITTED status. This situation can occur when the Workflow neither accepts nor rejects the Update due to:
- No available workers
- Workflow Task fails in the same tick

## Why?

Developers need to distinguish between `null` as an actual result of a Workflow Update action and "no result."
Prior to this patch, the code used `assert($result === null)`, which did not align with reality. Additionally, assertions are typically disabled in production environments.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/Client/Update/UpdateHandle.php
+++ b/src/Client/Update/UpdateHandle.php
@@ -13,6 +13,7 @@ use Temporal\DataConverter\ValuesInterface;
 use Temporal\Exception\Client\CanceledException;
 use Temporal\Exception\Client\TimeoutException;
 use Temporal\Exception\Client\WorkflowUpdateException;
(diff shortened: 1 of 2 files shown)


Sushisource:
@roxblnfk So I think we don't need to do this at all - in the other SDKs we just keep trying until we hit the timeout or are cancelled and we should mirror that behavior here.

wolfy-j (approved):

roxblnfk:
Replaced with #591

PR #527, opened 2024-11-15
Parameters for test server

cv65kr:
## What was changed
Possibility to add parameters for test server

## Why?
To have better configuration options

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/testing/src/Environment.php
+++ b/testing/src/Environment.php
@@ -50,7 +50,10 @@ final class Environment
         $this->startRoadRunner($rrCommand, $commandTimeout, $envs);
     }
 
-    public function startTemporalServer(int $commandTimeout = 10): void
+    /**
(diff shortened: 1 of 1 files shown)


testing/src/Environment.php
```diff
@@ -50,7 +50,10 @@ public function start(string $rrCommand = null, int $commandTimeout = 10, array
         $this->startRoadRunner($rrCommand, $commandTimeout, $envs);
     }
 
-    public function startTemporalServer(int $commandTimeout = 10): void
+    /**
+     * @param string[] $parameters
```
roxblnfk, line 54:
```suggestion
     * @param list<non-empty-string> $parameters
```

roxblnfk (approved):

PR #543, opened 2025-01-03
Add metadata for Meta Storm pluign

roxblnfk:
## What was changed

- Added metadata that covers a few client methods
- Added short advice in readme

image

> [!NOTE]
> It's just metadata for [the concrete plugin](https://github.com/xepozz/meta-storm-idea-plugin) and it doesn't affect anything.

## Why?

To have a better autocomplete

## Done

Client API
- [x] Autocomplete Workflow types in client API
	- [x] Fallback to short class name if WF type is not set
- [x] Autocomplete Workflow classes in client API

Workflow

- [x] Autocomplete Workflow and Activity classes in Workflow context


resources/.meta-storm.xml
```diff
@@ -0,0 +1,37 @@
...
+            class="\Temporal\Client\WorkflowClientInterface"
+            method="newWorkflowStub"
+            argument="0"
+        >
+            <collection name="workflow-class" argument="0" />
+        </classMethod>
+    </definitions>
+    <collections>
+        <attributeArgument
+            name="workflow-type"
```
xepozz, line 28:
```suggestion
            name="temporal/sdk:workflow-type"
```
Could you please use the prefix for the collections?

xepozz (changes requested):

msmakouz (approved):

PR #549, opened 2025-01-08
Use `namespace` from ClientOptions on creating of Schedule

### LINKED ISSUE #548: [Bug] Namespace from ClientOptions in ScheduleClient constructor not respected in createSchedule

I want to create a schedule in a specific Temporal namespace by specifying the namespace in the ScheduleClient constructor.

### Describe the bug

When I create a schedule without explicitly passing ScheduleOptions, the schedule is created in the default namespace instead of the namespace provided in ClientOptions during the ScheduleClient initialization.

This happens because of this two lines in the source code: [1](https://github.com/temporalio/sdk-php/blob/master/src/Client/ScheduleClient.php#L77), [2](https://github.com/temporalio/sdk-php/blob/master/src/Client/ScheduleClient.php#L84), where the namespace defaults to 'default'. I expect the namespace provided in the `ScheduleClient` constructor to be used as the default, similar to the behavior in [this line](https://github.com/temporalio/sdk-php/blob/master/src/Client/ScheduleClient.php#L139).

### Minimal Reproduction

```php
use Temporal\Client\ScheduleClient;
use Temporal\Client\GRPC\ServiceClient;
use Temporal\Client\ClientOptions;
use Temporal\Client\Schedule;

$serviceClient = new ServiceClient(new WorkflowServiceClient('localhost:7233', [
     'credentials' => ChannelCredentials::createInsecure(),
]));

$scheduleClient = new ScheduleClient(
    $serviceClient,
    (new ClientOptions())->withNamespace('my-namespace'),
);
$handle1 = $handle = $client->createSchedule(
    Schedule\Schedule::new(),
);
// $handle1->namespace contains 'default' 

$client->getHandle($handle1->getID())->trigger(); // throws an exception
$client->getHandle($handle1->getID(), 'default')->trigger(); // works fine

// just in case, this works as expected
$handle2 = $handle = $client->createSchedule(
    Schedule\Schedule::new(),
    options: (Schedule\ScheduleOptions::new())->withNamespace('my-namespace'),
);
// $handle2->namespace contains 'my-namespace'
$client->getHandle($handle2->getID())->trigger(); // works fine
``` 

### Expected Behavior
When `ScheduleClient` is initialized with `ClientOptions` specifying a namespace, that namespace should be used as the default for creating schedules.

### Environment/Versions

- OS and processor: **M2 Mac**
- Temporal Version: **Temporal 1.26.2, SDK 2.12.0**
- Are you using Docker or Kubernetes or building Temporal from source? **Docker**

roxblnfk:
## What was changed

- `ScheduleOptions->naemspace` has `null` value by default instead of `"default"`
- Deprecate `namespace` property and `withNamespace()` method in `ScheduleOptions`
- Use `namespace` from the ClientOptions as a fallback value when a Schedule is created

## Checklist

1. Closes #548
2. How was this tested: added unit tests

### DIFF
--- a/src/Client/Schedule/ScheduleOptions.php
+++ b/src/Client/Schedule/ScheduleOptions.php
@@ -16,20 +16,23 @@ final class ScheduleOptions
 {
     use CloneWith;
 
-    public readonly string $namespace;
-    public readonly bool $triggerImmediately;
+    /**
+     * @deprecated will be removed in the next major version.
+     */
+    public readonly ?string $namespace;
 
     /**
      * @var list<BackfillPeriod>
      */
     public readonly array $backfills;
 
+    public readonly bool $triggerImmediately;
     public readonly EncodedCollection $memo;
     public readonly EncodedCollection $searchAttributes;
 
     private function __construct()
     {
-        $this->namespace = 'default';
+        $this->namespace = null;
         $this->triggerImmediately = false;
         $this->backfills = [];
         $this->memo = EncodedCollection::empty();
@@ -41,6 +44,10 @@ final class ScheduleOptions
         return new self();
     }
 
+    /**
+     * @deprecated Configure the namespace on the {@see \Temporal\Client\ClientOptions} instead
+     *             when creating the {@see \Temporal\Client\ScheduleClient}.
+     */
     public function withNamespace(string $namespace): self
     {
         /** @see self::$namespace */
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -109,10 +109,21 @@
       <code><![CDATA[$workflowType]]></code>
     </PropertyNotSetInConstructor>
   </file>
+  <file src="src/Client/Schedule/ScheduleOptions.php">
+    <DeprecatedProperty>
+      <code><![CDATA[$this->namespace]]></code>
+    </DeprecatedProperty>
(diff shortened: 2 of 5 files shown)


src/Client/Schedule/ScheduleOptions.php
```diff
@@ -16,20 +16,23 @@ final class ScheduleOptions
 {
     use CloneWith;
 
-    public readonly string $namespace;
-    public readonly bool $triggerImmediately;
+    /**
+     * @deprecated will be removed in the next major version.
+     */
+    public readonly ?string $namespace;
```
AleksandrKosovS, line 22: Just as a note: [ScheduleClient::getHandle](https://github.com/temporalio/sdk-php/blob/master/src/Client/ScheduleClient.php#L131) still allows specifying a namespace as an argument and does not appear to be deprecated.

Currently, `createSchedule` creates schedules in the given namespace (while passing the namespace as an argument is marked as deprecated), whereas `getHandle` supports working with handles in different namespaces. Is this behavior intentional?

roxblnfk: It's hard to deprecate a parameter.
We will most likely remove this in the next major version.

wolfy-j (approved):

PR #581, opened 2025-03-11
Add methods to register fallback handlers

### LINKED ISSUE #508: [Feature Request] Add  fallback Update/Signal/Query listeners

Right now we declare handlers dynamically using only known names.
It would be nice to have the ability to declare a handler for signals, queries, and updates whose names we don't know in advance.

Add something like `Workflow::registerListener()` function to define fallback update/signal/query listeners

### Additional context

[Java docs](https://docs.temporal.io/develop/java/message-passing#set-a-dynamic-signal)

roxblnfk:
## What was changed

Added methods to define dynamic handlers for Signals, Updates, and Queries.
- `Workflow::registerDynamicQuery($handler)`
- `Workflow::registerDynamicSignal($handler)`
- `Workflow::registerDynamicUpdate($handler, $validator)`

## Checklist

1. Closes #508
3. Todo:
    - [x] check that naming is OK
    - [x] should we add the ability to unset the handlers?

### Dynamic Signal Handler

You can register a Signal handler that will be called if a handler for a specific Signal is not found.

If by the time the dynamic handler is registered, the Workflow has already received unhandled Signals, they will be processed immediately in the order they were received.

```php
 \Temporal\Workflow::registerDynamicSignal(function (string $name, ValuesInterface $arguments): void {
     \error_log(\sprintf(
         'Executed signal `%s` with %d arguments',
         $name,
         $arguments->count(),
     ));
 });
```

### Dynamic Query Handler

You can register a Query handler that will be called if a handler for a specific Query is not found.

```php
\Temporal\Workflow::registerDynamicQuery(function (string $name, ValuesInterface $arguments): string {
    return \sprintf(
        'Got query `%s` with %d arguments',
        $name,
        $arguments->count(),
    );
});
```

### Dynamic Update Handler

You can register an Update handler that will be called if a handler for a specific update is not found.
The `registerDynamicUpdate()` method accepts two arguments:
- Update handler.
- Update validator (optional) that should throw an exception if the validation fails.

```php
\Temporal\Workflow::registerDynamicUpdate(
    static fn(string $name, ValuesInterface $arguments): string => \sprintf(
        'Got update `%s` with %d arguments',
        $name,
        $arguments->count(),
    ),
    static fn(string $name, ValuesInterface $arguments) => \str_starts_with(
        $name,
        'update_',
    ) or throw new \InvalidArgumentException('Invalid update name'),
);
```


wolfy-j (approved):

Sushisource:
This all looks good to me but we tend to call these "dynamic" handlers in the other SDKs (though, I agree fallback probably makes more sense).

I think we probably should name them that just to be consistent

roxblnfk:
We already have methods like [`registerSignal(name, handler)`](https://github.com/temporalio/sdk-php/blob/ac9e85846f8d0fa6d4fc6db12ad62ae2962dd778/src/Workflow.php#L362), [`registerQuery(name, handler)`](https://github.com/temporalio/sdk-php/blob/ac9e85846f8d0fa6d4fc6db12ad62ae2962dd778/src/Workflow.php#L342), and [`registerUpdate(name, handler)`](https://github.com/temporalio/sdk-php/blob/ac9e85846f8d0fa6d4fc6db12ad62ae2962dd778/src/Workflow.php#L395) in this SDK.
As you can see, they don't have the `dynamic` suffix in the name, even though they work at runtime, and there's no option to unregister handlers.
That's why the questions arise.

PR #585, opened 2025-03-18
Support UserMetadata in Client API

roxblnfk:
## What was changed

- Added ExecutionConfig with UserMetadata in Workflow Description
- Added support for user metadata in Workflow Start/Schedule methods. Metadata in Timers and Activities require changes in RoadRunner and can be added in the future.

## Checklist

1. Part of #517

## Docs

Client API

```php
use Temporal\Client\GRPC\ServiceClient;
use Temporal\Client\ScheduleClient;
use Temporal\Client\Schedule\Action\StartWorkflowAction;
use Temporal\Client\WorkflowClient;
use Temporal\Client\WorkflowOptions;

$serviceClient = ServiceClient::create('127.0.0.1:7233');
$workflowClient = WorkflowClient::create($serviceClient);
$scheduleClient = ScheduleClient::create($serviceClient);

// Start Workflow with user metadata
$stub = $workflowClient->newUntypedWorkflowStub(
    'SimpleWorkflow',
    (new WorkflowOptions())
        ->withStaticSummary('some text')
        ->withStaticDetails('details') 
);
$workflowClient->start($stub);

// Describe workflow
echo $stub->describe()->config->userMetadata->summary;
echo $stub->describe()->config->userMetadata->details;

// Schedule Workflow with user metadata
$schedule = $scheduleClient->createSchedule(
    \Temporal\Client\Schedule\Schedule::new()
        ->withAction(StartWorkflowAction::new(SimpleWorkflow::class)
            ->withStaticSummary('some-summary')
            ->withStaticDetails('some-details'))
);

// Describe schedule
$action = $schedule->describe()->schedule->action;
assert($action instanceof StartWorkflowAction);

echo $action->userMetadata->details;
echo $action->userMetadata->summary;
```

Workflow context

```php
$stub = \Temporal\Workflow::newChildWorkflowStub(
    SimpleWorkflow::class,
    (new Workflow\ChildWorkflowOptions())
        ->withStaticSummary('some text')
        ->withStaticDetails('details')
);
```


src/Internal/Client/WorkflowStarter.php
```diff
@@ -335,7 +340,8 @@ private function configureExecutionRequest(
             ->setWorkflowTaskTimeout(DateInterval::toDuration($options->workflowTaskTimeout))
             ->setMemo($options->toMemo($this->converter))
             ->setSearchAttributes($options->toSearchAttributes($this->converter))
-            ->setHeader($header->toHeader());
+            ->setHeader($header->toHeader())
+            ->setUserMetadata($metadata);
```
Sushisource, line 344: You want to avoid setting it unless it actually contains something. Same goes for defaulting to empty strings. Proto will distinguish between unset metadata, and set metadata with empty fields. We don't want to actually set it unless something inside it has explicitly been set.

roxblnfk: What about Run/Execution/Task Timeouts? Is it better to skip zero Durations or set to `{seconds: 0, nanos:0}`?
I have already encountered instances where Temporal Cloud returns empty intervals as `null`, unlike the dev server. Should the SDK do the same now?

Sushisource: Generally speaking we should avoid setting any field unless the user explicitly configured it

Sushisource (approved):

PR #593, opened 2025-04-11
Expose Activity and Workflow instance from context

### LINKED ISSUE #546: [Feature Request] Access to workflow/activity instance from context

Provide access to the Temporal created/owned workflow and activity instance from their respective contexts. See https://github.com/temporalio/features/issues/572.

roxblnfk:
## What was changed

- Added `Activity::getInstance()` and `Workflow::getInstance()`
- Changed workflow execution flow:
    - First, the Workflow is initialized.
        - It is recommended not to make calls to start Activity, ChildWorkflow, Timer, etc.
        - If the `#[WorkflowInit]` attribute is present, the arguments are resolved.
    - `WorkflowInboundCallInterceptor::execute()` is called
        - Arguments from the previous step are used, but they can be overridden.
    - Workflow Handler is called.
- Now errors from `WorkflowInboundCallInterceptor::execute()` are recorded in the Workflow history.

Tests:
- Acceptance Tests: added ability to customize workers per TaskQueue in tests.
- Added tests for workflow dynamic handlers.

## Checklist

1. Closes #546
3. Any docs updates needed?

### DIFF
--- a/src/Internal/Workflow/Process/Process.php
+++ b/src/Internal/Workflow/Process/Process.php
@@ -23,7 +23,6 @@ use Temporal\Interceptor\WorkflowInbound\SignalInput;
 use Temporal\Interceptor\WorkflowInbound\UpdateInput;
 use Temporal\Interceptor\WorkflowInbound\WorkflowInput;
 use Temporal\Interceptor\WorkflowInboundCallsInterceptor;
-use Temporal\Internal\Declaration\MethodHandler;
 use Temporal\Internal\Declaration\WorkflowInstance;
 use Temporal\Internal\Declaration\WorkflowInstanceInterface;
 use Temporal\Internal\ServiceContainer;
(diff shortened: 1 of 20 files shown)


wolfy-j (approved):

src/Internal/Workflow/Process/Process.php
```diff
@@ -175,22 +173,68 @@ function (\Throwable $e): void {
...
+        $instance = $context->getWorkflowInstance();
+        $arguments = null;
 
-            // Prepare typed input
-            \assert($handler instanceof MethodHandler);
-            $arguments = $handler->resolveArguments($values);
+        try {
+            // Initialize workflow instance
+            //
+            // Resolve arguments if #[WorkflowInit] is used
```
Sushisource, line 190: Something we do in the other SDKs is we actually run the init method in a special "readonly" context that will throw errors if the user attempts to issue any commands during init. I dunno if that's something that you can readily enforce, but if possible it's a nice bit of safety.

Sushisource (approved):

src/Internal/Workflow/Process/Process.php
roxblnfk: Now any request from the constructor will throw an exception.
image

Workflow Context will no longer be readonly right after the init stage before the interceptor method `WorkflowInboundCallsInterceptor::execute` is run.

PR #620, opened 2025-05-19
Activity pause

### LINKED ISSUE #588: Heartbeating activities should be interrupted when the activities are paused.
See https://github.com/temporalio/features/issues/602

roxblnfk:
## What was changed

Added `ActivityPausedException` that is thrown on getting `paused` flag from heartbeat.

## Checklist

1. Closes #588

### DIFF
--- a/src/Exception/Client/ActivityPausedException.php
+++ b/src/Exception/Client/ActivityPausedException.php
@@ -1,0 +1,20 @@
+<?php
+
+/**
+ * This file is part of Temporal package.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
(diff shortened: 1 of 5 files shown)


wolfy-j (approved):

Sushisource:
I don't think you do, really. From SDK's perspective it's just a new activity attempt.

However, I do think we need to add some additional info to the context, since it's possible for multiple of these exceptional states to be happening, now. Ex, in Python: https://github.com/temporalio/sdk-python/blob/633b90c60580b73027053ca95becec0f37d09eeb/temporalio/activity.py#L147

Context of the whole commit here: https://github.com/temporalio/sdk-python/commit/27cc67f1c62b7c1ad058fa7ea2bee49ed061676e

Sushisource (approved):
Looking good!

PR #671, opened 2025-12-18
Maintenance: add Symfony 8

xepozz:
## What was changed

allowed symfony packages with 8.0 versions

## Why?

allow to use together with the latest symfony packages

## Checklist

1. Closes 670

2. How was this tested:

3. Any docs updates needed?

PR #674, opened 2025-12-19
chore: exclude temporal root file from git

xepozz:
## What was changed

ignore temporal binary from git index

## Why?

because it's downloaded into source root as well as test server, which is already excluded

https://github.com/temporalio/sdk-php/pull/669 was accidentally merged to another PR

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ Thumbs.db
 /.rr.yaml
 *.exe
 rr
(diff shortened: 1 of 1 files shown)


roxblnfk:
Just let's merge #671

PR #704, opened 2026-01-28
Remove docker compose using old server image

tconley1428:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,13 +58,14 @@
     "require-dev": {
         "buggregator/trap": "^1.13.0",
         "composer/composer": "^2.8.4",
+        "cweagans/composer-patches": "^2.0",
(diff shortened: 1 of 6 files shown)


xepozz:
> It wasn't clear to me that this docker-compose was still being used as it has no references in the repo that I could find. If it is still relevant, yes. If it is, it should be updated like this: [temporalio/sdk-core#1103](https://github.com/temporalio/sdk-core/pull/1103)

I remembered the one place: https://learn.temporal.io/getting_started/php/hello_world_in_php/, but it points to https://github.com/temporalio/samples-php/blob/master/docker-compose.yml, which also could be updated.

Maybe this compose file is also pointed from somewhere, but I'd add instructions how to run tests with docker instead of deleting it.

PR #724, opened 2026-03-06
feat: Plugins implementation

### LINKED ISSUE #644: [Feature Request] Plugin support

See https://github.com/temporalio/features/issues/652

xepozz:
## What was changed

- Added interfaces for the plugins
- Each interface has several methods that are called during the sdk working process
- Each interface has no-op trait to be able to implement only necessary methods
- There is `AbstractPlugin` instead of `SimplePlugin` as from the rest implementations
- Using plugins with the same name raise an error
- There's a convenient way to add plugins to a test class with php attribute
- Worker provides used plugins, but RoadRunner plugin doesn't support it yet

## Why?

Resolves https://github.com/temporalio/sdk-php/issues/644

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


src/Plugin/ConnectionPluginInterface.php
```diff
@@ -0,0 +1,36 @@
...
+     */
+    public function getName(): string;
+
+    /**
+     * Modify the service client before it is used by the client.
+     *
+     * Use this hook to configure connection-level settings such as
+     * API keys, gRPC metadata, auth tokens, or context options.
+     */
+    public function configureServiceClient(ConnectionPluginContext $context): void;
```
Sushisource: This doesn't do chain-of-responsibility like in the other SDKs - as a general rule I think we want all of the plugins to be able to work that way, as it allows for true interception rather than just configuration.

tests/Unit/Plugin/PluginPropagationTestCase.php
```diff
@@ -0,0 +1,239 @@
...
+
+        $factory = new WorkerFactory(
+            DataConverter::createDefault(),
+            $this->mockRpc(),
+            pluginRegistry: new PluginRegistry([$factoryPlugin]),
+            client: $client,
+        );
+        $factory->newWorker();
+
+        self::assertSame(['from-factory', 'from-client'], $order);
```
Sushisource: I believe we would want the client plugins to be applied first. The other SDKs are generally prepending the ones from the client.

Sushisource (commented):
General notes:
* Not seeing a way to configure the replay worker explicitly. This exists in the other SDKs.
* The new interfaces should be marked experimental

Sushisource (approved):
This is looking good to me but I didn't see any additional attention drawn to https://github.com/temporalio/sdk-php/pull/724#discussion_r2985075065 or maybe we can mitigate that somehow? Not a blocker in any case.

PR #735, opened 2026-03-25
fix: testing environment config file ignored and needlessly wait for timeout

mtorromeo:
## What was changed
Changed `Temporal\Testing\Environment::startRoadRunner` to actually use the `$configFile` parameter and to not wait for timeout if the process is not running.

## Why?
1. In `Temporal\Testing\Environment` the `startRoadRunner` has a `$configFile` parameter, which is also passed through the `start` method's `$roadRunnerConfigFile` parameter but that is completely ignored.

2. In `startRoadRunner` if the executed process crashes/exists before `$commandTimeout` triggers (which is 10 seconds by default but could also be much longer) then the function will wait needlessly for the timeout while it could return right away.

## Checklist
Confirmed by testing it on my own project, also with a misconfigured command that does not start correctly.


xepozz (approved):

mtorromeo:
Can this be merged anytime soon? Thanks

Issue #601, opened 2025-04-17
[Feature Request] Expose `WorkflowInboundCallInterceptor::init()`


roxblnfk:
Expose `WorkflowInboundCallInterceptor::init()` like in [Java SDK](https://github.com/temporalio/sdk-java/blob/a4c9e68254e4eae7e46ee0cd6c8f3e0d13dfe183/temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowInboundCallsInterceptor.java#L182-L194)

PR #739, opened 2026-04-11
feat: expose `WorkflowInboundCallInterceptor::init()`

xepozz:
## What was changed

- Added `WorkflowInboundCallInterceptor::init()` method
- `init` methods isn't the same as Java's one, it's a lifecycle stage: Workflow initing/end

## Why?

Closing #601


xepozz:
@roxblnfk please review if it's what you wanted

roxblnfk (commented):
It doesn't seem like the interceptor is running in the right place and providing functionality similar to that in the Java SDK.

xepozz:
> It doesn't seem like the interceptor is running in the right place and providing functionality similar to that in the Java SDK.

Indeed, so there are no need to re-implement Java way in PHP, isn't it?
It would be great if you add more details about the idea to the issue.

PR #768, opened 2026-06-23
feat: add Nexus integration

xepozz:
## What was changed

Samples: https://github.com/temporalio/samples-php/pull/75

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

## Related

- Features https://github.com/temporalio/features/pull/871
- Features PHP https://github.com/temporalio/features/pull/870
- Samples PHP https://github.com/temporalio/samples-php/pull/75
- SDK PHP https://github.com/temporalio/sdk-php/pull/768
- Roadrunner Temporal https://github.com/temporalio/roadrunner-temporal/pull/804


src/Workflow/NexusOperationOptions.php
```diff
@@ -0,0 +1,117 @@
...
+     * Service name to call. If empty, derived from the service interface.
+     */
+    #[Marshal(name: 'service')]
+    public string $service = '';
+
+    /**
+     * Overall timeout for the Nexus operation.
+     */
+    #[Marshal(name: 'scheduleToCloseTimeout', type: DateIntervalType::class)]
+    public \DateInterval $scheduleToCloseTimeout;
```
Quinn-With-Two-Ns, line 46: There should also be `scheduleToStart` and `startToCloseTimeout`

src/Nexus/WorkflowHandle.php
```diff
@@ -0,0 +1,41 @@
...
+declare(strict_types=1);
+
+namespace Temporal\Nexus;
+
+use Temporal\Client\WorkflowOptions;
+
+/**
+ * Description of the workflow that backs a Nexus operation.
+ */
+final class WorkflowHandle
```
VegetarianOrc, line 19: I was a little confused by the naming here as it felt a bit like this was a reference to a running workflow, but it's actually the input to a start workflow call.

src/Internal/Nexus/NexusTaskHandler.php
```diff
@@ -0,0 +1,252 @@
...
+     * Absolute deadline from Operation-Timeout (preferred) or Request-Timeout; absent or unparseable yields null.
+     *
+     * @param array<string, string> $headers
+     */
+    public static function deadlineFromHeaders(array $headers): ?\DateTimeImmutable
+    {
+        $lowerHeaders = \array_change_key_case($headers, \CASE_LOWER);
+
+        $value = NexusHeader::get($lowerHeaders, NexusHeader::OPERATION_TIMEOUT)
+            ?? NexusHeader::get($lowerHeaders, NexusHeader::REQUEST_TIMEOUT);
```
VegetarianOrc: These timeouts are separate and shouldn't be combined in this way. The request timeout is used to indicate how long the Nexus handler has to act. The operation timeout is how long the entire thing has to complete including any async activity started by the handler.

We should parse request timeout only here and leave operation timeout in the headers that are passed to the handler.

src/Internal/Nexus/NexusTaskHandler.php
```diff
@@ -0,0 +1,252 @@
...
+        foreach ($request->getHeader() as $key => $value) {
+            $headers[(string) $key] = (string) $value;
+        }
+
+        $context = new OperationContext(
+            service: $cancelRequest->getService(),
+            operation: $cancelRequest->getOperation(),
+            headers: $headers,
+            env: $this->env,
+        );
```
VegetarianOrc, line 193: The context here also needs the request timeout parsed as the deadline.

src/Nexus/WorkflowHandle.php
xepozz: This mirrors java's [io.temporal.nexus.WorkflowHandle](https://github.com/temporalio/sdk-java/blob/master/temporal-sdk/src/main/java/io/temporal/nexus/WorkflowHandle.java) naming
If you think that we should rename it, please suggest alternatives

src/Internal/Nexus/NexusTaskHandler.php
xepozz: Done

src/Workflow/NexusOperationOptions.php
xepozz: Added

PR #774, opened 2026-07-10
Support dynamic (catch-all) workflows

datashaman:
## What

Adds **Dynamic (catch-all) Workflow** support — a workflow invoked when the worker receives a type name that is not statically registered. PHP and TypeScript are the only SDKs without it (Go/Java/Python/.NET/Ruby all have it).

- **`#[WorkflowMethod(dynamic: true)]`** — declares the dynamic workflow. The handler reads the real type via `Workflow::getInfo()` and receives the raw args.
- **`WorkflowPrototype::isDynamic()`** — set by `WorkflowReader` from the attribute.
- **`StartWorkflow`** — when `workflows->find($type)` misses, falls back to the registered dynamic prototype (mirrors Python's `self._workflows.get(type, self._dynamic_workflow)`).
- **`GetWorkerInfo`** — advertises `dynamic` per workflow, so the RoadRunner temporal plugin can register a Go dynamic-workflow proxy.
- **`WorkflowCollection`** — enforces **at most one** dynamic workflow per worker (Go panics, Python raises `TypeError`; here a `LogicException`), since a second catch-all would make dispatch ambiguous.

## Why

Enables applications that author workflows at runtime (e.g. UI-driven pipeline/automation builders) to give each workflow its own type name — real identity in the Web UI — while one PHP handler interprets it, with no codegen or per-workflow deploy.

## Integration status

The host-side dependencies are complete:

- [temporalio/sdk-go#2449](https://github.com/temporalio/sdk-go/pull/2449) fixed factory-registered dynamic workflow execution and shipped in `go.temporal.io/sdk v1.47.0`.
- [temporalio/roadrunner-temporal#784](https://github.com/temporalio/roadrunner-temporal/pull/784) registers the shared factory proxy via `RegisterDynamicWorkflow` when this `dynamic` field is set. It merged on August 6, 2026, and pins Go SDK v1.47.0. The change is not yet included in a tagged `roadrunner-temporal` release or released RoadRunner binary.

This PR (the PHP declaration + dispatch) remains self-contained and unit-tested independently of the RoadRunner release cycle.

## Testing

- New `tests/Unit/Declaration` coverage, run against **both** the attribute and annotation readers:
  - `testDynamicWorkflow` — `#[WorkflowMethod(dynamic: true)]` sets `isDynamic()`.
  - `testMultipleDynamicWorkflowsAreRejected` — a second dynamic workflow on a worker throws.
- Full unit suite green (`composer run test:unit`, 692 tests). `composer run cs:diff` clean. `composer run psalm` reports no new issues in the changed files.
- Validated end-to-end with a custom `rr` containing the now-merged roadrunner-temporal#784 implementation and Go SDK v1.47.0: starting an unregistered type (`pipeline-blog-publish`) is caught by the dynamic handler, which reads the real type via `Workflow::getInfo()` and completes.

Automated end-to-end coverage can use the standard test binary once a RoadRunner release includes the merged plugin change.

## Backwards compatibility

Additive. `dynamic` defaults to `false`, so existing workflows and workers behave exactly as before.

### DIFF
--- a/src/Workflow/WorkflowMethod.php
+++ b/src/Workflow/WorkflowMethod.php
@@ -36,10 +36,22 @@ final class WorkflowMethod
     public ?string $name = null;
 
     /**
+     * Marks this as a dynamic (catch-all) workflow: it is invoked when the
+     * worker receives a workflow whose type name is not statically registered.
+     * At most one dynamic workflow may be registered per worker. The handler
+     * reads the actual type name via {@see \Temporal\Workflow::getInfo()} and
+     * receives the raw arguments (declare a {@see \Temporal\DataConverter\ValuesInterface}
+     * parameter to access them).
+     */
+    #[Immutable]
+    public bool $dynamic = false;
+
+    /**
      * @param non-empty-string|null $name
      */
-    public function __construct(?string $name = null)
+    public function __construct(?string $name = null, bool $dynamic = false)
     {
         $this->name = $name;
+        $this->dynamic = $dynamic;
     }
 }
--- a/tests/Unit/Declaration/Fixture/WorkflowWithAnotherDynamic.php
+++ b/tests/Unit/Declaration/Fixture/WorkflowWithAnotherDynamic.php
@@ -1,0 +1,26 @@
+<?php
+
+/**
+ * This file is part of Temporal package.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Temporal\Tests\Unit\Declaration\Fixture;
+
+use Temporal\Workflow\WorkflowInterface;
+use Temporal\Workflow\WorkflowMethod;
+
+/** @WorkflowInterface */
+#[WorkflowInterface]
+class WorkflowWithAnotherDynamic
+{
+    /** @WorkflowMethod(dynamic=true) */
+    #[WorkflowMethod(dynamic: true)]
+    public function handler(): void
+    {
+    }
+}
--- a/tests/Unit/Declaration/Fixture/WorkflowWithDynamic.php
+++ b/tests/Unit/Declaration/Fixture/WorkflowWithDynamic.php
@@ -1,0 +1,26 @@
+<?php
+
+/**
+ * This file is part of Temporal package.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Temporal\Tests\Unit\Declaration\Fixture;
+
+use Temporal\Workflow\WorkflowInterface;
+use Temporal\Workflow\WorkflowMethod;
+
+/** @WorkflowInterface */
+#[WorkflowInterface]
+class WorkflowWithDynamic
+{
+    /** @WorkflowMethod(dynamic=true) */
+    #[WorkflowMethod(dynamic: true)]
+    public function handler(): void
+    {
+    }
+}
--- a/src/Internal/Declaration/Prototype/WorkflowCollection.php
+++ b/src/Internal/Declaration/Prototype/WorkflowCollection.php
@@ -12,8 +12,37 @@ declare(strict_types=1);
 namespace Temporal\Internal\Declaration\Prototype;
 
 use Temporal\Internal\Repository\ArrayRepository;
+use Temporal\Internal\Repository\Identifiable;
 
 /**
  * @template-extends ArrayRepository<WorkflowPrototype>
  */
-final class WorkflowCollection extends ArrayRepository {}
+final class WorkflowCollection extends ArrayRepository
+{
+    /**
+     * A dynamic (catch-all) workflow is the single fallback used when no
+     * statically registered workflow matches the requested type. As in the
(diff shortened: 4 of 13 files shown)


datashaman:
The RoadRunner side is now merged in temporalio/roadrunner-temporal#784, including the required Go SDK v1.47.0 update.

I’ve also added the missing PHP-side runtime and regression coverage here: raw `ValuesInterface` arguments for dynamic handlers, preservation of the actual workflow type, named-workflow precedence, per-worker registration, handshake metadata, and the no-fallback failure case. The full unit suite passes (1,185 tests / 3,397 assertions).

This PR is ready for review and can be merged independently. The only remaining dependency is a new roadrunner-temporal/RoadRunner release containing #784 before the feature can be used with an official binary; end-to-end behavior has already been validated with a custom build.

Please let me know if there's anything else I can add.

src/Workflow/WorkflowMethod.php
```diff
@@ -35,11 +35,23 @@ final class WorkflowMethod
     #[Immutable]
     public ?string $name = null;
 
+    /**
+     * Marks this as a dynamic (catch-all) workflow: it is invoked when the
+     * worker receives a workflow whose type name is not statically registered.
+     * At most one dynamic workflow may be registered per worker. The handler
+     * reads the actual type name via {@see \Temporal\Workflow::getInfo()} and
+     * receives the raw arguments (declare a {@see \Temporal\DataConverter\ValuesInterface}
+     * parameter to access them).
```
xepozz, line 44:
```suggestion
     * Marks this as a dynamic (catch-all) workflow: it is invoked when the
     * worker receives a workflow whose type name is not statically registered.
     * At most one dynamic workflow may be registered per worker.
```

tests/Unit/Declaration/Fixture/WorkflowWithAnotherDynamic.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Temporal\Tests\Unit\Declaration\Fixture;
+
+use Temporal\Workflow\WorkflowInterface;
+use Temporal\Workflow\WorkflowMethod;
+
+/** @WorkflowInterface */
```
xepozz, line 17:
```suggestion
```

tests/Unit/Declaration/Fixture/WorkflowWithAnotherDynamic.php
```diff
@@ -0,0 +1,26 @@
...
+#[WorkflowInterface]
+class WorkflowWithAnotherDynamic
+{
+    /** @WorkflowMethod(dynamic=true) */
```
xepozz, line 21:
```suggestion
```

tests/Unit/Declaration/Fixture/WorkflowWithDynamic.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Temporal\Tests\Unit\Declaration\Fixture;
+
+use Temporal\Workflow\WorkflowInterface;
+use Temporal\Workflow\WorkflowMethod;
+
+/** @WorkflowInterface */
```
xepozz, line 17:
```suggestion
```

tests/Unit/Declaration/Fixture/WorkflowWithDynamic.php
```diff
@@ -0,0 +1,26 @@
...
+#[WorkflowInterface]
+class WorkflowWithDynamic
+{
+    /** @WorkflowMethod(dynamic=true) */
```
xepozz, line 21:
```suggestion
```

src/Workflow/WorkflowMethod.php
datashaman: Applied in 52c4d4f9 — shortened the documentation to the suggested description.

tests/Unit/Declaration/Fixture/WorkflowWithAnotherDynamic.php
datashaman: Applied in 52c4d4f9 — removed the legacy WorkflowInterface annotation from this fixture.

tests/Unit/Declaration/Fixture/WorkflowWithAnotherDynamic.php
datashaman: Applied in 52c4d4f9 — removed the legacy WorkflowMethod annotation. I also scoped the dynamic-workflow tests to AttributeReader so the attribute-only fixture is tested coherently.

tests/Unit/Declaration/Fixture/WorkflowWithDynamic.php
datashaman: Applied in 52c4d4f9 — removed the legacy WorkflowInterface annotation from this fixture.

tests/Unit/Declaration/Fixture/WorkflowWithDynamic.php
datashaman: Applied in 52c4d4f9 — removed the legacy WorkflowMethod annotation. The affected tests now use AttributeReader directly.

PR #776, opened 2026-07-14
Defer schedule catchup window default to server

chaptersix:
## What changed
- Make the default catch-up window nullable instead of assigning 60 seconds.
- Remove null policy fields during protobuf mapping.
- Preserve exact protobuf duration values when decoding server responses.
- Correct the public documentation.
- Add omission, exact-duration decoding, and server-backed one-year default coverage.

## Why
The SDK was sending a historical 60-second default, preventing the server’s one-year default from taking effect. Its protobuf decoder also converted long fixed durations through a calendar diff, which changed the reported total seconds.

## Validation
- GitHub Actions static analysis
- Unit coverage for exact protobuf duration decoding
- Server-backed schedule coverage confirms the 365-day default
- Tests were not run locally because PHP and Composer are unavailable.

## Breaking changes
`SchedulePolicies::$catchupWindow` is nullable until the schedule is resolved by the server.

### DIFF
--- a/src/Client/Schedule/Policy/SchedulePolicies.php
+++ b/src/Client/Schedule/Policy/SchedulePolicies.php
@@ -6,6 +6,7 @@ namespace Temporal\Client\Schedule\Policy;
 
 use Google\Protobuf\Duration;
 use Temporal\Internal\Marshaller\Meta\Marshal;
+use Temporal\Internal\Marshaller\Type\DateIntervalType;
 use Temporal\Internal\Support\DateInterval;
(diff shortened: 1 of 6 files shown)


chaptersix:
I'm not a PHP dev so apologies if this is slop. We just need to correct this docs and allow the server to select the proper default.

chaptersix:
important thing is that the default should be 1 year. if this sdk wants to continue setting it, that's fine.

xepozz:
Is this affects you somehow or you just want to allow sdk to use server defaults?

PR #778, opened 2026-07-20
feat: mock activities and child workflows in the testing framework

### LINKED ISSUE #302: [Feature Request] Set the expectCompletion result for each Activity call inside the loop

For PHP-SDK we can use `ActivityMocker` and `expectCompletion()` to mock an activity call result. And it works great if we don't call the activity inside the loop.

If we call it inside the loop than we can't set `expectCompletion()` for each of the Activity call inside the loop.

For example consider we have the following phpunit test:

```php
public function testWorkflowSend()
{
    $workflow = $this->workflowClient->newWorkflowStub(SomeWorkflow::class);
    $run = $this->workflowClient->start($workflow, ['value1', 'value2', 'value3']);
    $this->assertSame(['someResult1', 'someResult2', 'someResult3'], $run->getResult('array'));
}
```
The workflow method:
```php
public function send(array $values)
{
    $results = [];
    foreach ($values as $value) {
        $results[] = yield $this->activity->doSomething($value);
    }
    return $results;
}
```
And we want to mock the activity calls like this:
```php
$this->activityMocks->expectCompletion(
    'SomeActivity.doSomething', ['someResult1', 'someResult2', 'someResult3']
);
```

The result of running the workflow inside this test will be:
```php
array(3) {
[0] => array(3) {
    [0] => string(11) "someResult1",
    [1] => string(11) "someResult2",
    [2] => string(11) "someResult3",
}, 
[1] => array(3) {
    [0] => string(11) "someResult1",
    [1] => string(11) "someResult2",
    [2] => string(11) "someResult3",
}, 
[2] => array(3) {
    [0] => string(11) "someResult1",
    [1] => string(11) "someResult2",
    [2] => string(11) "someResult3",
}
```
Instead of:
```php
array(9) {
    [0] => string(11) "someResult1",
    [1] => string(11) "someResult2",
    [2] => string(11) "someResult3",
}
```
Even if we mock the activity calls like this:
```php
$this->activityMocks->expectCompletion('SomeActivity.doSomething', 'someResult1');
$this->activityMocks->expectCompletion('SomeActivity.doSomething', 'someResult2');
$this->activityMocks->expectCompletion('SomeActivity.doSomething', 'someResult3');
```
The result will be:
```php
array(9) {
    [0] => string(11) "someResult3",
    [1] => string(11) "someResult3",
    [2] => string(11) "someResult3",
}
```

It will be great if we have a choice to set `expectCompletion` result for each call of the Activity inside the loop. For example:

```php
$this->activityMocks->expectCompletion(
    'SomeActivity.doSomething', 'someResult1', 1
); // for the first call

$this->activityMocks->expectCompletion(
    'SomeActivity.doSomething', 'someResult2', 2
); // for the second

$this->activityMocks->expectCompletion(
    'SomeActivity.doSomething', 'someResult3', 3
); // for the third
```
And the same with `expectFailure()`:
```php
$this->activityMocks->expectCompletion(
    'SomeActivity.doSomething', 'someResult1', 1
); // for the first call
$this->activityMocks->expectFailure(
    'SomeActivity.doSomething', new Error(), 2
); // except error for the second call
```

### LINKED ISSUE #524: [Feature Request] Allow mocking child workflows like activites

In a situation where one workflow starts a child workflow, there is no way to replace the child workflow with a stub. Or I don't understand how to do it - there is no word about it in the documentation.

I found some strange behavior that led me to these thoughts.
Let's imagine that we have two workflows:

1. Workflow A
2. Workflow B, which is run via `asyncDetached` in Workflow A. Workflow B has a timer for 30 minutes

This tricky setup of two workflows allows to detect a bug (as I think). If we run an integration test on Workflow A where only the activity are replaced with stubs, then run an integration test on Workflow B, the timer waits an honest 30 minutes.

If we run only the test on Workflow B, the timer will run instantly without any waiting.

It seems that if you replace Workflow A in the test with a stub, this strange problem goes away.

xepozz:
## What was changed

- `WorkflowMocker` — stub child workflows (`expectCompletion` / `expectFailure` / `expectCompletionWhen`) without starting a real child execution, plus `expectVersion` and `expectSideEffect`.
- `ActivityMocker` — per-call results via `expectConsecutiveCompletions` and arg-matched `expectCompletionWhen`.
- `WorkflowInteractions` — history-based assertions over a finished run: activity, local activity, child workflow, timer and external signal calls (call counts + input matching).
- `WorkflowTestCase` base test case wiring the mocks and interactions together.

## Why?

Bring the PHP testing framework closer to Go/Java/TS: child workflows were the only outbound call with no way to stub it, and activity mocks couldn't return a different value per call inside a loop.

## Checklist

1. Closes #524,
   closes #302

3. How was this tested:

New Unit + Functional suites: child/activity mocking (incl. failure cause preservation and arg matching), `WorkflowInteractions`, `getVersion`/`sideEffect` mocks, and a time-skipping case proving a stubbed 30-min-timer child fast-forwards in < 60s wall-clock.

4. Any docs updates needed?

Yes — document child-workflow mocking and `WorkflowInteractions` on docs.temporal.io (PHP testing-suite page).

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1118,17 +1118,6 @@
       <code><![CDATA[$reduce($c, $value, $i++, $total)]]></code>
     </TooManyArguments>
   </file>
-  <file src="src/Worker/ActivityInvocationCache/ActivityInvocationResult.php">
-    <MissingReturnType>
-      <code><![CDATA[toValue]]></code>
-    </MissingReturnType>
-    <PossiblyNullArgument>
-      <code><![CDATA[$dataConverter]]></code>
-    </PossiblyNullArgument>
-    <PossiblyUndefinedStringArrayOffset>
-      <code><![CDATA[$data['payloads']]]></code>
-    </PossiblyUndefinedStringArrayOffset>
-  </file>
   <file src="src/Worker/ActivityInvocationCache/InMemoryActivityInvocationCache.php">
     <PossiblyUndefinedStringArrayOffset>
       <code><![CDATA[$request->getOptions()['name']]]></code>
@@ -1141,6 +1130,20 @@
     </ArgumentTypeCoercion>
     <PossiblyUndefinedStringArrayOffset>
       <code><![CDATA[$request->getOptions()['name']]]></code>
+    </PossiblyUndefinedStringArrayOffset>
+  </file>
+  <file src="src/Worker/ChildWorkflowInvocationCache/RoadRunnerChildWorkflowInvocationCache.php">
+    <ArgumentTypeCoercion>
+      <code><![CDATA[$cacheName]]></code>
+      <code><![CDATA[$host]]></code>
+    </ArgumentTypeCoercion>
+  </file>
+  <file src="src/Worker/InvocationResult.php">
+    <PossiblyNullArgument>
+      <code><![CDATA[$dataConverter]]></code>
+    </PossiblyNullArgument>
+    <PossiblyUndefinedStringArrayOffset>
+      <code><![CDATA[$data['payloads']]]></code>
     </PossiblyUndefinedStringArrayOffset>
   </file>
   <file src="src/Worker/LoopInterface.php">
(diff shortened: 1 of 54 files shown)


wolfy-j (approved):

igancev:
Wonderful, thanks! 🙏🎉

